Issue #45, opened 2024-04-29
DEV -  Updating the environment variable isSecret value through terraform issues errors


shri-astro:
**Describe the bug**

The following resource description was used to update the deployment's environment variable isSecret value from false to true retaining rest of the deployment details intact.
The error message received is mentioned in this ticket

Deployment update resource definition:
```
resource "astro_deployment" "hybrid" {
  name                  = "my hybrid deployment"
  description           = "an example deployment"
  type                  = "HYBRID"
  cluster_id            = "clnp86ly5000401ndagu20g81"
  task_pod_node_pool_id = "clnp86ly5000301ndzfxz895w"
  contact_emails        = ["example@astronomer.io"]
  executor              = "KUBERNETES"
  is_cicd_enforced      = false
  is_dag_deploy_enabled = false
  scheduler_replicas    = 1
  scheduler_au          = 5
  workspace_id          = "cl0v1p6lc728255byzyfs7lw21"
  environment_variables = [{
    key       = "key1"
    value     = "value1"
    is_secret = true
  }]
}
```
Error message:
```
astro_deployment.hybrid: Modifying... [id=clvl0m9pd03gj01jnm9ki469n]
╷
│ Error: Provider produced inconsistent result after apply
│ 
│ When applying changes to astro_deployment.hybrid, provider "provider[\"registry.terraform.io/astronomer/astro\"]" produced an unexpected new value: .environment_variables: inconsistent values for
│ sensitive attribute.
│ 
│ This is a bug in the provider, which should be reported in the provider's own issue tracker.

```

**What Terraform Provider Version and Terraform Version did you experience this bug?**

Latest version of the terraform provider
Terraform v1.8.2

vandyliu:
should be fixed

PR #60, opened 2024-05-14
Fix deployment env vars not working if is_secret=True

vandyliu:
## Description
- Previously, environment variables with `is_secret=True` emit errors in Terraform
- On create, update, and read for deployment resources, we must give Terraform the actual secret value so it doesnt error out and see inconsistencies. The `value` field is sensitive though so it is not outputted, however it is necessary so Terraform doesn't complain and error out.

## 🎟 Issue(s)
https://github.com/astronomer/terraform-provider-astro/issues/45
## 🧪 Functional Testing
- Added acceptance integration tests
- Also works with imported deployment

## 📋 Checklist

- [x] Rebased from the main (or release if patching) branch (before testing)
- [x] Tested against [Astro-API](https://github.com/astronomer/astro/)


frank-ye (approved):

internal/provider/resources/resource_deployment.go
```diff
@@ -91,8 +91,35 @@ func (r *DeploymentResource) Create(
...
+		return
+	}
+	if deploymentOptions.JSON200 == nil || len(deploymentOptions.JSON200.RuntimeReleases) == 0 {
+		resp.Diagnostics.AddError(
+			"Client Error",
+			"Unable to get runtime releases for deployment creation, got empty runtime releases",
+		)
+		return
+	}
+
```
vandyliu, line 119:
```suggestion
```

Issue #61, opened 2024-05-14
Cannot Create Deployment With Scaling_Spec.Hibernation_Spec without hibernation_spec.override


ichung08:
**Describe the bug**

Cannot create a standard deployment that uses `scaling_spec.hibernation_spec.schedules` without `hibernation_spec.override`. From the docs, it says that both `override` and `schedules` is omit empty but when you try to create a deployment with _just_ `schedules` we get the following error:

```
╷
│ Error: Value Conversion Error
│ 
│   with astro_deployment.team_1_dev_deployment,
│   on main.tf line 21, in resource "astro_deployment" "team_1_dev_deployment":
│   21: resource "astro_deployment" "team_1_dev_deployment" {
│ 
│ An unexpected error was encountered trying to build a value. This is always an error in the provider. Please report the following to the provider developer:
│ 
│ Received null value, however the target type cannot handle null values. Use the corresponding `types` package type, a pointer type or a custom type that handles null values.
│ 
│ Path: hibernation_spec.override
│ Target Type: models.HibernationSpecOverride
│ Suggested `types` Type: basetypes.ObjectValue
│ Suggested Pointer Type: *models.HibernationSpecOverride

```

Creating a deployment with `override` and no `schedules` works.

**What Terraform Provider Version and Terraform Version did you experience this bug?**
Latest

**What Operating System is the above Terraform Provider installed on?**
macOS

**🪜 Steps To Reproduce**

Run `terraform apply` with the following script in `main.tf`

```

resource "astro_workspace" "team_1_workspace" {
  name                  = "Team 1 Workspace"
  description           = "Team 1 Workspace"
  cicd_enforced_default = true
}


resource "astro_deployment" "team_1_dev_deployment" {
  name                    = "Team 1 Dev Deployment"
  description             = "Team 1 Dev Deployment"
  type                    = "STANDARD"
  cloud_provider          = "AWS"
  region                  = "us-east-1"
  contact_emails          = []
  default_task_pod_cpu    = "0.25"
  default_task_pod_memory = "0.5Gi"
  executor                = "CELERY"
  is_cicd_enforced        = true
  is_dag_deploy_enabled   = true
  is_development_mode     = true
  is_high_availability    = false
  resource_quota_cpu      = "10"
  resource_quota_memory   = "20Gi"
  scheduler_size          = "SMALL"
  workspace_id            = astro_workspace.team_1_workspace.id
  environment_variables   = []
  worker_queues = [{
[...]

PR #62, opened 2024-05-15
Fix creating and updating hosted deployments with scaling specs

vandyliu:
## Description
- This PR fixes issues where you cannot create a deployment with a scaling spec
- Terraform previously would error out

## 🎟 Issue(s)
https://github.com/astronomer/terraform-provider-astro/issues/61
## 🧪 Functional Testing
- Added lots of tests for scaling spec

## 📋 Checklist

- [x] Rebased from the main (or release if patching) branch (before testing)
- [x] Tested against [Astro-API](https://github.com/astronomer/astro/)


internal/provider/models/deployment.go
```diff
@@ -303,54 +295,112 @@ type HibernationSchedule struct {
...
+func HibernationOverrideTypesObject(
+	ctx context.Context,
+	hibernationOverride *platform.DeploymentHibernationOverride,
+) (types.Object, diag.Diagnostics) {
+	if hibernationOverride == nil {
+		return types.ObjectNull(schemas.HibernationOverrideAttributeTypes()), nil
+	}
+	obj := HibernationSpecOverride{
+		IsHibernating: types.BoolPointerValue(hibernationOverride.IsHibernating),
+		IsActive:      types.BoolPointerValue(hibernationOverride.IsActive),
```
ianbuss, line 324: Quick note: `IsActive` is a property set on the API response, it's not something that can be set on the request.

ianbuss (commented):
Don't know the tf API well enough to approve this but it looks sane to me and I like the code style @vandyliu 👍

internal/provider/models/deployment.go
vandyliu: Yep! got confused a bit by this sometimes but this is for creating the "Terraform response" similar to other "computed" values we have like `updatedAt`
In resource_deployment.go, we create the request where `isActive` is not there

internal/provider/resources/resource_deployment.go
```diff
@@ -118,7 +118,7 @@ func (r *DeploymentResource) Create(
 		return
 	}
 
-	var diags diag.Diagnostics
+	diags := make(diag.Diagnostics, 0)
```
frank-ye, line 121: For my own understanding - should these all be `[]diag.Diagnostics`? Or is the object itself an array?

frank-ye: I guess the initialization value of 0 is just the "null" val here?

vandyliu: `diag.Diagnostics` is equivalent to `[]diag.Diagnostic`

vandyliu: so here its making an empty array

internal/provider/models/deployment.go
```diff
@@ -244,23 +246,13 @@ func (data *DeploymentResource) ReadFromResponse(
 	data.SchedulerSize = types.StringPointerValue((*string)(deployment.SchedulerSize))
 	data.IsDevelopmentMode = types.BoolPointerValue(deployment.IsDevelopmentMode)
 	data.IsHighAvailability = types.BoolPointerValue(deployment.IsHighAvailability)
-
-	// Currently, the scaling status and spec are only available in development mode
-	// However, there is a bug in the API where the scaling status and spec are returned even if the deployment is not in development mode for updated deployments
-	// This is a workaround to handle the bug until the API is fixed
-	// Issue here: https://github.com/astronomer/astro/issues/21073
```
vandyliu, line 251: issue is fixed

internal/provider/models/deployment.go
```diff
@@ -459,11 +441,11 @@ func WorkerQueueDataSourceTypesObject(
 }
 
 type DeploymentScalingSpec struct {
-	HibernationSpec HibernationSpec `tfsdk:"hibernation_spec"`
```
vandyliu, line 462: had to change it to use types.Object for resource creation since it would error otherwise

fredzhy (approved):

internal/provider/resources/resource_deployment_test.go
```diff
@@ -282,7 +283,7 @@ func TestAcc_ResourceDeploymentStandard(t *testing.T) {
 				ResourceName:            azureCeleryResourceVar,
 				ImportState:             true,
 				ImportStateVerify:       true,
-				ImportStateVerifyIgnore: []string{"external_ips", "oidc_issuer_url", "environment_variables.1.value"}, // environment_variables.0.value is a secret value
+				ImportStateVerifyIgnore: []string{"external_ips", "oidc_issuer_url", "scaling_status", "environment_variables.1.value"}, // environment_variables.0.value is a secret value
```
frank-ye, line 286: Sanity check: `ImportStateVerifyIgnore` means that TF won't return this in the output of `terraform plan` or `terraform apply`? How do these affect the assertions?

internal/provider/resources/resource_deployment_test.go
```diff
@@ -324,7 +325,228 @@ func TestAcc_ResourceDeploymentStandard(t *testing.T) {
...
+					Name:              scalingSpecDeploymentName,
+					Description:       utils.TestResourceDescription,
+					Region:            "us-east4",
+					CloudProvider:     "GCP",
+					Executor:          "CELERY",
+					SchedulerSize:     "SMALL",
+					IsDevelopmentMode: true,
+					ScalingSpec:       `scaling_spec = {}`,
+				}),
+				ExpectError: regexp.MustCompile(`Inappropriate value for attribute "scaling_spec"`),
```
frank-ye, line 356: nit: `invalid` over `inappropriate` is more accurate here

internal/provider/resources/resource_deployment_test.go
```diff
@@ -324,7 +325,228 @@ func TestAcc_ResourceDeploymentStandard(t *testing.T) {
...
+					Executor:          "CELERY",
+					SchedulerSize:     "SMALL",
+					IsDevelopmentMode: true,
+					ScalingSpec: `
+						scaling_spec = {
+							hibernation_spec = {
+								schedules = []
+							}
+						}`}),
+				ExpectError: regexp.MustCompile(`Invalid Attribute Value`), // schedules must have at least one element
```
frank-ye, line 424: nit - not a blocker, any way we can ensure the assertion on the right attribute here? i.e. `schedules` here

internal/provider/resources/resource_deployment_test.go
```diff
@@ -324,7 +325,228 @@ func TestAcc_ResourceDeploymentStandard(t *testing.T) {
...
+								override = {
+								  is_hibernating = true
+								  override_until = "2075-01-01T00:00:00Z"
+								}
+							}
+						}`}),
+				Check: resource.ComposeTestCheckFunc(
+					resource.TestCheckResourceAttr(scalingSpecResourceVar, "scaling_spec.hibernation_spec.override.is_hibernating", "true"),
+					resource.TestCheckResourceAttr(scalingSpecResourceVar, "scaling_spec.hibernation_spec.override.override_until", "2075-01-01T00:00:00Z"),
+					resource.TestCheckNoResourceAttr(scalingSpecResourceVar, "scaling_spec.hibernation_spec.schedules"),
```
frank-ye, line 510: nit: for each config, can add comments in assertions on what we're checking - or maybe in the description field something like "Scaling Spec with Hibernation Spec Override Without Schedule" idk if it's overkill, but just an idea for test maintainability

internal/provider/resources/resource_deployment_test.go
vandyliu: It means that for this test case, we are just ignore those values after an import - we will not check if they are equivalent
Since the first 3 values are computed, it doesnt really matter that the values are not the same and the last value cannot be the same since its a secret
I had it this way since it was throwing errors for those 3

internal/provider/resources/resource_deployment_test.go
vandyliu: inappropriate is what the terraform error outputs through their library

internal/provider/resources/resource_deployment_test.go
vandyliu: so I refactored the tests to make it a lil more obvious as to what they do
I think I will skip adding comments just in case we update the tests but forget to update the comments but now the tests should be a lot simpler and readable and the main gist of what is going on should be extremely obvious
Also the tests aren't necessarily for asserting the responses but more importantly to check that it doesnt error or panic when it should succeed

PR #83, opened 2024-06-05
Update README.md

sunkickr:
## Description

I'm under the impression that you need to have a version specified for this example to work. Want to make sure it works as is so I added a version

## 📋 Checklist

- [ ] Rebased from the main (or release if patching) branch (before testing)
- [ ] Tested against [Astro-API](https://github.com/astronomer/astro/)

### DIFF
--- a/README.md
+++ b/README.md
@@ -83,6 +83,7 @@ terraform {
   required_providers {
     astro = {
       source = "registry.terraform.io/astronomer/astro"
+      version = "0.2.0"
     }
   }
 }


README.md
```diff
@@ -83,6 +83,7 @@ terraform {
   required_providers {
     astro = {
       source = "registry.terraform.io/astronomer/astro"
+      version = "0.2.0"
```
vandyliu, line 86: lets do this so that when we release new versions, we dont have to update it
```suggestion
      version >= "0.0.0"
```

README.md
```diff
@@ -83,6 +83,7 @@ terraform {
   required_providers {
     astro = {
       source = "registry.terraform.io/astronomer/astro"
```
vandyliu, line 85:
```suggestion
      source = "astronomer/astro"
```

vandyliu: we should update this in all our other doc files too

README.md
```diff
@@ -40,6 +40,7 @@ terraform {
   required_providers {
     astro = {
       source = "astronomer/astro"
+      version >= "0.0.0"
```
vandyliu, line 43: you don't need a version
without version, it gets the latest one already

sunkickr: Iv had customers cpmplain that this is not the case

vandyliu: are they using the one on the registry?
i just tested it and it worked fine without version on a brand new terraform project

sunkickr: maybe they were doing something else wrong. I haven't heard this complaint in awhile

Issue #105, opened 2024-07-09
Add api tokens resource


ichung08:

vandyliu:
things we will need to think about
- token roles inside the resource (this resource might need to make up to 2 calls in the create/update)
- rotating the token

PR #110, opened 2024-07-11
Add API Token Resource

ichung08:
## Description

Add API Token resource to terraform. 

Supports:
- Creating an API Token
- Updating an API Token (name, description and roles)
- Deleting an API Token

## 🎟 Issue(s)

#105 

## 🧪 Functional Testing

**Create**
Screenshot 2024-07-11 at 2 04 49 PM
Screenshot 2024-07-11 at 2 04 58 PM

**Update**
Screenshot 2024-07-15 at 2 53 48 PM

**Delete**
Screenshot 2024-07-11 at 2 38 09 PM

## 📋 Checklist


examples/resources/astro_api_token/resource.tf
```diff
@@ -0,0 +1,11 @@
+resource "astro_api_token" "example" {
```
vandyliu, line 1: we should add some examples for 
- org tokens with deployment and workspace roles
- workspace token
- workspace token with deployment role
- deployment token with `DEPLOYMENT_ADMIN` role
- deployment token with custom role

internal/provider/models/api_token.go
```diff
@@ -67,3 +86,48 @@ func (data *ApiTokenDataSource) ReadFromResponse(ctx context.Context, apiToken *
...
 	return diags
 }
+
+func (data *ApiTokenResource) ReadFromResponse(ctx context.Context, apiToken *iam.ApiToken) diag.Diagnostics {
+	var diags diag.Diagnostics
+	data.Id = types.StringValue(apiToken.Id)
+	data.Name = types.StringValue(apiToken.Name)
+	if apiToken.Description != "" {
+		data.Description = types.StringValue(apiToken.Description)
+	}
```
vandyliu, line 96: I feel like this would make more sense
```suggestion
	data.Description = types.StringValue(apiToken.Description)
```

vandyliu: or maybe this since description is `optional`
```suggestion
	if apiToken.Description == "" {
		data.Description = types.StringNull()
	} else {
	        data.Description = types.StringValue(apiToken.Description)
	}
```

regardless we should add test cases where we create a token with empty description then update it to non-empty and vice versa case

internal/provider/schemas/api_token.go
```diff
@@ -72,3 +75,76 @@ func ApiTokenDataSourceSchemaAttributes() map[string]datasourceSchema.Attribute
...
+			MarkdownDescription: "API Token description",
+			Optional:            true,
+		},
+		"short_token": resourceSchema.StringAttribute{
+			MarkdownDescription: "API Token short token",
+			Computed:            true,
+		},
+		"type": resourceSchema.StringAttribute{
+			MarkdownDescription: "API Token type",
+			Required:            true,
```
vandyliu, line 102: we should have stringvalidator here with `oneof`

internal/provider/schemas/api_token.go
```diff
@@ -72,3 +75,76 @@ func ApiTokenDataSourceSchemaAttributes() map[string]datasourceSchema.Attribute
...
+		"last_used_at": resourceSchema.StringAttribute{
+			MarkdownDescription: "API Token last used timestamp",
+			Computed:            true,
+		},
+		"roles": resourceSchema.SetNestedAttribute{
+			NestedObject: resourceSchema.NestedAttributeObject{
+				Attributes: ResourceApiTokenRoleSchemaAttributes(),
+			},
+			Required:            true,
+			MarkdownDescription: "The roles assigned to the API Token",
```
vandyliu, line 143: we should have validator at least one

internal/provider/schemas/api_token.go
```diff
@@ -72,3 +75,76 @@ func ApiTokenDataSourceSchemaAttributes() map[string]datasourceSchema.Attribute
...
+		},
+		"token": resourceSchema.StringAttribute{
+			MarkdownDescription: "API Token value",
+			Computed:            true,
```
vandyliu, line 147: this should have `Sensitive: true`

vandyliu: since its an authorization token and should be used carefully

internal/provider/schemas/role.go
```diff
@@ -103,6 +103,36 @@ func ApiTokenRoleAttributeTypes() map[string]attr.Type {
 	}
 }
 
+func ResourceApiTokenRoleSchemaAttributes() map[string]resourceSchema.Attribute {
```
vandyliu, line 106: if you look at `resource_deployment.go`'s `ValidateConfig`
We should have something like that for the roles.

For example, if its an organization token, then it can have any role
If it is a deployment token, it can just have one role
In this we should also check that the role matches the entity. Ie. workspace roles for workspace entities.

vandyliu: we should also add a bunch of error test cases for these in the acceptance tests

internal/provider/models/api_token.go
```diff
@@ -67,3 +86,48 @@ func (data *ApiTokenDataSource) ReadFromResponse(ctx context.Context, apiToken *
...
+		data.LastUsedAt = types.StringValue("")
+	}
+	data.Roles, diags = utils.ObjectSet(ctx, apiToken.Roles, schemas.ApiTokenRoleAttributeTypes(), ApiTokenRoleTypesObject)
+	if diags.HasError() {
+		return diags
+	}
+	if apiToken.Token != nil {
+		data.Token = types.StringValue(*apiToken.Token)
+	} else {
+		data.Token = types.StringValue("")
```
vandyliu, line 130: not sure how this will work since when you do anything other than create, the token value will not appear

internal/provider/resources/resource_api_token.go
```diff
@@ -0,0 +1,339 @@
...
+	roles, diags := RequestApiTokenRoles(ctx, data.Roles)
+	if diags.HasError() {
+		resp.Diagnostics.Append(diags...)
+		return
+	}
+
+	// Create the API token request
+	createApiTokenRequest := iam.CreateApiTokenRequest{
+		Name: data.Name.ValueString(),
+		Role: roles[0].Role,
```
vandyliu, line 102: no guarantee that this will be the correct role
we should filter for the role depending on the token type and entity type of roles

internal/provider/resources/resource_api_token.go
```diff
@@ -0,0 +1,339 @@
...
+	// If the entity type is WORKSPACE or DEPLOYMENT, set the entity id
+	if createApiTokenRequest.Type == iam.WORKSPACE || createApiTokenRequest.Type == iam.DEPLOYMENT {
+		createApiTokenRequest.EntityId = lo.ToPtr(roles[0].EntityId)
+	}
+
+	if !data.Description.IsNull() {
+		createApiTokenRequest.Description = data.Description.ValueStringPointer()
+	} else {
+		createApiTokenRequest.Description = lo.ToPtr("")
+	}
```
vandyliu, line 115: this way usually reads easier
```suggestion
	if data.Description.IsNull() {
		createApiTokenRequest.Description = lo.ToPtr("")
	} else {
		createApiTokenRequest.Description = data.Description.ValueStringPointer()
	}
```

internal/provider/resources/resource_api_token.go
```diff
@@ -0,0 +1,339 @@
...
+		resp.Diagnostics.Append(diagnostic)
+		return
+	}
+
+	diags = data.ReadFromResponse(ctx, apiToken.JSON200)
+	if diags.HasError() {
+		resp.Diagnostics.Append(diags...)
+		return
+	}
+
```
vandyliu, line 145: we also need to update the roles here since there may be multiple roles

vandyliu: so if len(roles) > 1, you have to call the `UpdateApiTokenRoles` endpoint

internal/provider/resources/resource_api_token.go
```diff
@@ -0,0 +1,339 @@
...
+	// Read Terraform plan data into the model
+	resp.Diagnostics.Append(req.Plan.Get(ctx, &data)...)
+
+	if resp.Diagnostics.HasError() {
+		return
+	}
+
+	// update request
+	updateApiTokenRequest := iam.UpdateApiTokenJSONRequestBody{
```
vandyliu, line 220: here as well, 
prob best to call the update roles here first

internal/provider/models/api_token.go
ichung08: tested it with the second code snippet and it works fine, but it also works fine without explicitly setting the value to null

examples/resources/astro_api_token/resource.tf
```diff
@@ -0,0 +1,59 @@
...
+    "role" : "WORKSPACE_OWNER",
+    "entity_id" : "clx42sxw501gl01o0gjenthnh",
+    "entity_type" : "WORKSPACE"
+    },
+    {
+      "role" : "DEPLOYMENT_ADMIN",
+      "entity_id" : "clyn6kxud003x01mtxmccegnh",
+      "entity_type" : "DEPLOYMENT"
+  }]
+}
```
vandyliu, line 59: still need to add deployment token with custom role and with `DEPLOYMENT_ADMIN` role

internal/provider/models/api_token.go
```diff
@@ -67,3 +86,50 @@ func (data *ApiTokenDataSource) ReadFromResponse(ctx context.Context, apiToken *
...
+	}
+	data.Roles, diags = utils.ObjectSet(ctx, apiToken.Roles, schemas.ApiTokenRoleAttributeTypes(), ApiTokenRoleTypesObject)
+	if diags.HasError() {
+		return diags
+	}
+	if apiToken.Token == nil {
+		data.Token = types.StringNull()
+	} else {
+		data.Token = types.StringValue(*apiToken.Token)
+	}
```
vandyliu, line 133: we should make this consistent
either the first case is `== nil` or `!= nil` but shouldn't mix them

internal/provider/provider_test_utils.go
```diff
@@ -36,6 +36,7 @@ func TestAccPreCheck(t *testing.T) {
 		"HOSTED_USER_ID",
 		"HOSTED_WORKSPACE_ID",
 		"HOSTED_DEPLOYMENT_ID",
+		"HOSTED_API_TOKEN_ID",
```
vandyliu, line 39: don't think we need this

internal/provider/resources/resource_api_token.go
```diff
@@ -0,0 +1,566 @@
...
+			entityType = iam.ApiTokenRoleEntityTypeWORKSPACE
+		} else if v.EntityType.ValueString() == string(iam.ApiTokenRoleEntityTypeDEPLOYMENT) {
+			entityType = iam.ApiTokenRoleEntityTypeDEPLOYMENT
+		}
+
+		return iam.ApiTokenRole{
+			Role:       v.Role.ValueString(),
+			EntityId:   v.EntityId.ValueString(),
+			EntityType: entityType,
+		}
```
vandyliu, line 548:
```suggestion
	apiTokenRoles := lo.Map(roles, func(v models.ApiTokenRole, _ int) iam.ApiTokenRole {
		return iam.ApiTokenRole{
			Role:       v.Role.ValueString(),
			EntityId:   v.EntityId.ValueString(),
			EntityType: iam.ApiTokenRoleEntityType(v.EntityType.ValueString()),
		}
	})
```

internal/provider/resources/resource_api_token.go
```diff
@@ -0,0 +1,566 @@
...
+	})
+
+	return apiTokenRoles, nil
+}
+
+func RequestApiTokenRole(roles []iam.ApiTokenRole, entityType string) (string, string, diag.Diagnostics) {
```
vandyliu, line 554: let's return `iam.ApiTokenRole` instead

internal/provider/resources/resource_api_token.go
```diff
@@ -0,0 +1,566 @@
...
+		),
+	}
+}
+
+func validateDeploymentApiToken(roles []iam.ApiTokenRole) diag.Diagnostics {
+	for _, role := range roles {
+		if role.EntityType != iam.ApiTokenRoleEntityTypeDEPLOYMENT {
+			return diag.Diagnostics{
+				diag.NewErrorDiagnostic(
+					"Client Error",
```
vandyliu, line 504:
```suggestion
					"Bad Request Error",
```

internal/provider/resources/resource_api_token.go
```diff
@@ -0,0 +1,566 @@
...
+
+func validateOrganizationApiToken(roles []iam.ApiTokenRole) diag.Diagnostics {
+	for _, role := range roles {
+		if utils.CheckRole(role.Role, string(role.EntityType)) {
+			return nil
+		}
+	}
+	return diag.Diagnostics{
+		diag.NewErrorDiagnostic(
+			"Client Error",
```
vandyliu, line 470:
```suggestion
			"Bad Request Error",
```

internal/provider/resources/resource_api_token.go
```diff
@@ -0,0 +1,566 @@
...
+			"Unable to find the role for the entity type",
```
vandyliu, line 471:
```suggestion
			"There is no organization role in 'roles'",
```

internal/provider/resources/resource_api_token.go
```diff
@@ -0,0 +1,566 @@
...
+		),
+	}
+}
+
+func validateWorkspaceApiToken(roles []iam.ApiTokenRole) diag.Diagnostics {
+	for _, role := range roles {
+		if role.EntityType == iam.ApiTokenRoleEntityTypeORGANIZATION {
+			return diag.Diagnostics{
+				diag.NewErrorDiagnostic(
+					"Client Error",
```
vandyliu, line 481: All of these should be `Bad Request Error` with proper descriptions
`Client Error` is when the API call to core errors

internal/provider/resources/resource_api_token.go
```diff
@@ -0,0 +1,566 @@
...
+		if role.EntityType != iam.ApiTokenRoleEntityTypeDEPLOYMENT {
+			return diag.Diagnostics{
+				diag.NewErrorDiagnostic(
+					"Client Error",
+					"API Token of type DEPLOYMENT cannot have an ORGANIZATION or WORKSPACE role",
+				),
+			}
+		}
+
+		if utils.CheckRole(role.Role, string(role.EntityType)) {
```
vandyliu, line 510: we shouldn't be using utils/test in code here
if we want to use it, we should move it out
we should also rename this function to be more explicit, something like `ValidateRoleMatchesEntityType`

vandyliu: u could prob just call this in `ValidateConfig` since all three helpers use it the same way

internal/provider/resources/resource_api_token.go
```diff
@@ -0,0 +1,566 @@
...
+	case string(iam.ApiTokenRoleEntityTypeORGANIZATION):
+		resp.Diagnostics.Append(validateOrganizationApiToken(roles)...)
+	case string(iam.ApiTokenRoleEntityTypeWORKSPACE):
+		resp.Diagnostics.Append(validateWorkspaceApiToken(roles)...)
+	case string(iam.ApiTokenRoleEntityTypeDEPLOYMENT):
+		resp.Diagnostics.Append(validateDeploymentApiToken(roles)...)
+	}
+}
+
+func validateOrganizationApiToken(roles []iam.ApiTokenRole) diag.Diagnostics {
```
vandyliu, line 462: here we should check that there is org role and there is exactly 1

internal/provider/resources/resource_api_token.go
```diff
@@ -0,0 +1,566 @@
...
+	return diag.Diagnostics{
+		diag.NewErrorDiagnostic(
+			"Client Error",
+			"Unable to find the role for the entity type",
+		),
+	}
+}
+
+func validateDeploymentApiToken(roles []iam.ApiTokenRole) diag.Diagnostics {
+	for _, role := range roles {
```
vandyliu, line 500: here we should check that there is exactly one role and it is a deployment role and no other roles

internal/provider/resources/resource_api_token.go
```diff
@@ -0,0 +1,566 @@
...
+	if err != nil {
+		tflog.Error(ctx, "failed to create API token", map[string]interface{}{"error": err})
+		resp.Diagnostics.AddError(
+			"Client Error",
+			fmt.Sprintf("Unable to create API token and get API token, got error: %s", err),
+		)
+		return
+	}
+
+	// Set the token in the response
```
vandyliu, line 197:
```suggestion
	// Set the token in the response since it won't be returned in the GET call
```

internal/provider/resources/resource_api_token.go
```diff
@@ -0,0 +1,566 @@
...
+			return
+		}
+		_, diagnostic = clients.NormalizeAPIError(ctx, updatedApiToken.HTTPResponse, updatedApiToken.Body)
+		if diagnostic != nil {
+			resp.Diagnostics.Append(diagnostic)
+			return
+		}
+	}
+
+	// Get api token
```
vandyliu, line 182:
```suggestion
	// Get api token and use this as data since it will have the correct roles
```

internal/provider/resources/resource_api_token.go
```diff
@@ -0,0 +1,566 @@
...
+		)
+		return
+	}
+	_, diagnostic = clients.NormalizeAPIError(ctx, apiToken.HTTPResponse, apiToken.Body)
+	if diagnostic != nil {
+		resp.Diagnostics.Append(diagnostic)
+		return
+	}
+
+	// Get api token
```
vandyliu, line 342:
```suggestion
	// Get api token and use this as data since it will have the correct roles
```

internal/provider/resources/resource_api_token.go
```diff
@@ -0,0 +1,566 @@
...
+	if err != nil {
+		tflog.Error(ctx, "failed to update API token", map[string]interface{}{"error": err})
+		resp.Diagnostics.AddError(
+			"Client Error",
+			fmt.Sprintf("Unable to update API token and get API token, got error: %s", err),
+		)
+		return
+	}
+
+	diags = data.ReadFromResponse(ctx, apiTokenResp.JSON200)
```
vandyliu, line 357: you may need to get the `token` from `req.State` and add it to `apiTokenResp.JSON200`
but i think changing this function to 
```suggestion
	diags = data.ReadFromResponse(ctx, apiTokenResp.JSON200, token)
```
might be better
not sure how good of an idea it is to edit the response of an API - we might mess up somewhere where we don't want to

vandyliu: also would need to be done in `Read`

internal/provider/resources/resource_api_token.go
ichung08: good point - wasn't aware we could just do it like that

internal/provider/resources/resource_api_token.go
ichung08: do you mean `resp.State`? `req` doesn't have state as a field

also would the values set in create persist in read and update? (im assuming yes because the resource would be managed by tf and it would store all the values associated with it?)

vandyliu: nope i mean `req.State` 
its used in line 220

and yes

vandyliu: Screenshot 2024-07-18 at 2 24 40 PM

ichung08: wouldn't doing `data.Token.ValueString()` also work here? `req.State.Get` populates the prior state data into the target model (which would be the data model we define earlier)

vandyliu: sure something like `diags = data.ReadFromResponse(ctx, apiTokenResp.JSON200, data.Token.ValueString())` works

ichung08: in that case we won't need the previous line that edits the response output in create? or do we want both still

vandyliu: yea lets just do it this way`diags = data.ReadFromResponse(ctx, apiTokenResp.JSON200, data.Token.ValueString())` for `Create` `Read` `Update`

ichung08: i think create can be `*apiToken.JSON200.Token` since the `data.Token` wouldn't be populated with the new token output. i just pushed my new code so you can see

vandyliu: yep yep that looks correct

docs/resources/api_token.md
```diff
@@ -72,6 +72,33 @@ resource "astro_api_token" "example_workspace_token_with_deployment_role" {
...
+  roles = [{
+    "role" : "DEPLOYMENT_ADMIN",
+    "entity_id" : "clyn6kxud003x01mtxmccegnh",
+    "entity_type" : "DEPLOYMENT"
+    },
+    {
+      "role" : "CUSTOM_ROLE",
+      "entity_id" : "clyn6kxud003x01mtxmccegnh",
+      "entity_type" : "DEPLOYMENT"
+  }]
```
vandyliu, line 100: deployment tokens can only have one role

examples/resources/astro_api_token/resource.tf
```diff
@@ -56,4 +56,31 @@ resource "astro_api_token" "example_workspace_token_with_deployment_role" {
...
+
+resource "astro_api_token" "example_deployment_token_with_custom_role" {
+  name        = "deployment api token with custom role"
+  description = "deployment api token description"
+  type        = "DEPLOYMENT"
+  roles = [{
+    "role" : "DEPLOYMENT_ADMIN",
+    "entity_id" : "clyn6kxud003x01mtxmccegnh",
+    "entity_type" : "DEPLOYMENT"
+    },
```
vandyliu, line 80: can remove this since it can only have one role

internal/utils/api_token.go
```diff
@@ -0,0 +1,25 @@
+package utils
```
vandyliu, line 1: maybe this should just be called `role.go`
will prob apply to users

examples/resources/astro_api_token/resource.tf
ichung08: won't it still need `entity_id` and `entity_type`? that's also how the type is setup so those are required

vandyliu: ya its from the custom  role

vandyliu: like just have the one custom role that has all 3 fields but roles array should be length 1

examples/resources/astro_api_token/resource.tf
```diff
@@ -0,0 +1,81 @@
...
+  type        = "ORGANIZATION"
+  roles = [{
+    "role" : "ORGANIZATION_OWNER",
+    "entity_id" : "clx42kkcm01fo01o06agtmshg",
+    "entity_type" : "ORGANIZATION"
+  }]
+  expiry_period_in_days = 30
+}
+
+resource "astro_api_token" "example" {
```
vandyliu, line 13: lets have these two have different terraform variable names instead of both being `"example"`

internal/provider/schemas/api_token.go
```diff
@@ -72,3 +78,89 @@ func ApiTokenDataSourceSchemaAttributes() map[string]datasourceSchema.Attribute
...
+				Attributes: ResourceApiTokenRoleSchemaAttributes(),
+			},
+			Required:            true,
+			MarkdownDescription: "The roles assigned to the API Token",
+			Validators: []validator.Set{
+				setvalidator.SizeAtLeast(1),
+			},
+		},
+		"token": resourceSchema.StringAttribute{
+			MarkdownDescription: "API Token value",
```
vandyliu, line 161:
```suggestion
			MarkdownDescription: "API Token value. Warning: This value will be saved in plaintext in the terraform state file.",
```

internal/provider/resources/resource_api_token.go
```diff
@@ -0,0 +1,548 @@
...
+
+func RequestApiTokenRole(roles []iam.ApiTokenRole, entityType string) (iam.ApiTokenRole, diag.Diagnostics) {
+	for _, role := range roles {
+		if role.EntityType == iam.ApiTokenRoleEntityType(entityType) {
+			return role, nil
+		}
+	}
+	return iam.ApiTokenRole{}, diag.Diagnostics{
+		diag.NewErrorDiagnostic(
+			"Client Error",
```
vandyliu, line 544:
```suggestion
			"Bad Request Error",
```

internal/provider/resources/resource_api_token.go
```diff
@@ -0,0 +1,548 @@
...
+			"Unable to find the role for the entity type",
```
vandyliu, line 545: should have a more clear description for how user can fix this err

internal/provider/resources/resource_api_token.go
```diff
@@ -0,0 +1,548 @@
...
+	}
+
+	// If the entity type is WORKSPACE or DEPLOYMENT, set the entity id
+	if createApiTokenRequest.Type == iam.WORKSPACE || createApiTokenRequest.Type == iam.DEPLOYMENT {
+		role, diags = RequestApiTokenRole(roles, data.Type.ValueString())
+		if diags != nil {
+			resp.Diagnostics.Append(diags...)
+			return
+		}
+
```
vandyliu, line 121: i don't think we need this code since its already done in line 100
```suggestion
```

internal/provider/resources/resource_api_token.go
```diff
@@ -0,0 +1,548 @@
...
+			Role:       v.Role.ValueString(),
+			EntityId:   v.EntityId.ValueString(),
+			EntityType: iam.ApiTokenRoleEntityType(v.EntityType.ValueString()),
+		}
+	})
+
+	return apiTokenRoles, nil
+}
+
+func RequestApiTokenRole(roles []iam.ApiTokenRole, entityType string) (iam.ApiTokenRole, diag.Diagnostics) {
```
vandyliu, line 536:
```suggestion
func RequestApiTokenPrimaryRole(roles []iam.ApiTokenRole, entityType string) (iam.ApiTokenRole, diag.Diagnostics) {
```

internal/provider/resources/resource_api_token.go
```diff
@@ -0,0 +1,548 @@
...
+		if entityType == string(iam.ApiTokenRoleEntityTypeWORKSPACE) && role.EntityType == iam.ApiTokenRoleEntityTypeORGANIZATION {
+			return diag.Diagnostics{
+				diag.NewErrorDiagnostic(
+					"Bad Request Error",
+					"API Token of type WORKSPACE cannot have an ORGANIZATION role",
+				),
+			}
+		}
+
+		if entityType == string(iam.ApiTokenRoleEntityTypeDEPLOYMENT) && (role.EntityType != iam.ApiTokenRoleEntityTypeDEPLOYMENT || role.Role != "DEPLOYMENT_ADMIN") {
```
vandyliu, line 463:
```suggestion
		if entityType == string(iam.ApiTokenRoleEntityTypeDEPLOYMENT) && role.EntityType != iam.ApiTokenRoleEntityTypeDEPLOYMENT {
```

internal/provider/resources/resource_api_token.go
```diff
@@ -0,0 +1,548 @@
...
+			return diag.Diagnostics{
+				diag.NewErrorDiagnostic(
+					"Bad Request Error",
+					"API Token of type DEPLOYMENT cannot have an ORGANIZATION or WORKSPACE role and the role must be 'DEPLOYMENT_ADMIN'",
```
vandyliu, line 467:
```suggestion
					"API Token of type DEPLOYMENT cannot have an ORGANIZATION or WORKSPACE role",
```

vandyliu: it can have custom role

internal/provider/resources/resource_api_token_test.go
```diff
@@ -0,0 +1,568 @@
...
+		CheckDestroy: resource.ComposeTestCheckFunc(
+			// Check that the organization api token has been removed
+			testAccCheckApiTokenExistence(t, checkApiTokensExistenceInput{name: apiTokenName, organization: true, shouldExist: false}),
+		),
+		Steps: []resource.TestStep{
+			{
+				Config: astronomerprovider.ProviderConfig(t, true) + apiToken(apiTokenInput{
+					Name:        apiTokenName,
+					Description: utils.TestResourceDescription,
+					Type:        "ORGANIZATION",
```
vandyliu, line 45: nit we should use `iam.ORGANIZATION` for constants like this and the roles as well
```suggestion
					Type:        "ORGANIZATION",
```

internal/provider/resources/resource_api_token_test.go
```diff
@@ -0,0 +1,568 @@
...
+							EntityId:   workspaceId,
+							EntityType: "WORKSPACE",
+						},
+						{
+							Role:       "DEPLOYMENT_ADMIN",
+							EntityId:   deploymentId,
+							EntityType: "DEPLOYMENT",
+						},
+					},
+					ExpiryPeriodInDays: 30,
```
vandyliu, line 170:
```suggestion
```

internal/provider/resources/resource_api_token_test.go
```diff
@@ -0,0 +1,568 @@
...
+					resource.TestCheckResourceAttr(resourceVar, "roles.0.role", "WORKSPACE_OWNER"),
+					// Check via API that api token was destroyed and recreated
+					testAccCheckApiTokenExistence(t, checkApiTokensExistenceInput{name: apiTokenName, workspace: true, shouldExist: true}),
+				),
+			},
+			// Change resource type back to ORGANIZATION
+			{
+				Config: astronomerprovider.ProviderConfig(t, true) + apiToken(apiTokenInput{
+					Name:        apiTokenName,
+					Description: utils.TestResourceDescription,
```
vandyliu, line 151: lets remove the optional fields for this test and add that to the comment above and see if it works
```suggestion
```

internal/provider/resources/resource_api_token_test.go
```diff
@@ -0,0 +1,568 @@
...
+					Type:        "ORGANIZATION",
+					Roles: []apiTokenRole{
+						{
+							Role:       "ORGANIZATION_OWNER",
+							EntityId:   organizationId,
+							EntityType: "ORGANIZATION",
+						},
+					},
+					ExpiryPeriodInDays: 30,
```
vandyliu, line 289: same for this test
```suggestion
					Type:        "ORGANIZATION",
					Roles: []apiTokenRole{
						{
							Role:       "ORGANIZATION_OWNER",
							EntityId:   organizationId,
							EntityType: "ORGANIZATION",
						},
					},
```

internal/provider/resources/resource_api_token_test.go
vandyliu, line 1: lets also add some negative test cases for each token, where we have some bad inputs and ensure that there is an error - ie. the stuff we have in ValidateConfig

example search for `ExpectError` in `resource_deployment_test.go`

internal/provider/resources/resource_api_token_test.go
ichung08: `ExpiryPeriodInDays` cannot be removed after a value has been set previous - this aligns with the API as you can't update that field

tested in acc tests and you get:
```
        Invalid field values: field tokenExpiryPeriodInDays failed check on length;
        tokenExpiryPeriodInDays should be min: 1;, status: 400, requestId:
        bea9e0d6-7aa6-4cf8-b158-78cbb6316b80
```

internal/provider/models/api_token.go
```diff
@@ -67,3 +86,52 @@ func (data *ApiTokenDataSource) ReadFromResponse(ctx context.Context, apiToken *
...
+	if apiToken.LastUsedAt == nil {
+		data.LastUsedAt = types.StringValue("")
+	} else {
+		data.LastUsedAt = types.StringValue(apiToken.LastUsedAt.String())
+	}
+	data.Roles, diags = utils.ObjectSet(ctx, apiToken.Roles, schemas.ApiTokenRoleAttributeTypes(), ApiTokenRoleTypesObject)
+	if diags.HasError() {
+		return diags
+	}
+	if apiToken.Token != nil {
```
vandyliu, line 129:
```suggestion
	if apiToken.Token != nil && len(*apiToken.Token) > 0 {
```

internal/provider/resources/resource_api_token.go
```diff
@@ -0,0 +1,547 @@
...
+		updatedApiToken, err := r.IamClient.UpdateApiTokenRolesWithResponse(
+			ctx,
+			r.OrganizationId,
+			tokenId,
+			updateApiTokenRolesRequest,
+		)
+		if err != nil {
+			tflog.Error(ctx, "failed to create API token", map[string]interface{}{"error": err})
+			resp.Diagnostics.AddError(
+				"Bad Request Error",
```
vandyliu, line 163:
```suggestion
				"Client Error",
```

internal/provider/resources/resource_api_token.go
```diff
@@ -0,0 +1,547 @@
...
+
+	apiToken, err := r.IamClient.CreateApiTokenWithResponse(
+		ctx,
+		r.OrganizationId,
+		createApiTokenRequest,
+	)
+	if err != nil {
+		tflog.Error(ctx, "failed to create API token", map[string]interface{}{"error": err})
+		resp.Diagnostics.AddError(
+			"Bad Request Error",
```
vandyliu, line 137:
```suggestion
			"Client Error",
```

internal/provider/resources/resource_api_token.go
```diff
@@ -0,0 +1,547 @@
...
+			),
+		}
+	}
+
+	return diag.Diagnostics{
+		diag.NewErrorDiagnostic(
+			invalidRoleError,
+			fmt.Sprintf("Please provide a valid role for the entity type '%s'", entityType),
+		),
+	}
```
vandyliu, line 510: nit, usually we would want the `return nil` at the very end

internal/provider/resources/resource_api_token_test.go
```diff
@@ -0,0 +1,716 @@
...
+			},
+			// Test invalid role for entity type
+			{
+				Config: astronomerprovider.ProviderConfig(t, true) + apiToken(apiTokenInput{
+					Name: apiTokenName,
+					Type: string(iam.ORGANIZATION),
+					Roles: []apiTokenRole{
+						{
+							Role:       string(iam.WORKSPACEOWNER),
+							EntityId:   workspaceId,
```
vandyliu, line 65:
```suggestion
							EntityId:   organizationId,
```

internal/provider/resources/resource_api_token_test.go
```diff
@@ -0,0 +1,716 @@
...
+			},
+			// Test multiple roles of the same type
+			{
+				Config: astronomerprovider.ProviderConfig(t, true) + apiToken(apiTokenInput{
+					Name: apiTokenName,
+					Type: string(iam.ORGANIZATION),
+					Roles: []apiTokenRole{
+						{
+							Role:       string(iam.ORGANIZATIONOWNER),
+							EntityId:   workspaceId,
```
vandyliu, line 80:
```suggestion
							EntityId:   organizationId,
```

internal/provider/resources/resource_api_token_test.go
```diff
@@ -0,0 +1,716 @@
...
+							EntityType: string(iam.ORGANIZATION),
+						},
+						{
+							Role:       string(iam.ORGANIZATIONBILLINGADMIN),
+							EntityId:   workspaceId,
```
vandyliu, line 85:
```suggestion
							EntityId:   organizationId,
```

internal/provider/resources/resource_api_token.go
```diff
@@ -0,0 +1,547 @@
...
+		return
+	}
+}
+
+func validateApiTokenRoles(entityType string, roles []iam.ApiTokenRole) diag.Diagnostics {
+	var numRolesMatchingEntityType int
+	var invalidRoleError string
+
+	for _, role := range roles {
+		if entityType == string(iam.ApiTokenRoleEntityTypeWORKSPACE) && role.EntityType == iam.ApiTokenRoleEntityTypeORGANIZATION {
```
vandyliu, line 453: can we check that if `entityType` is organization then the `organizationId` matches `r.OrganizationId` and add a test case as well for when it doesnt

internal/provider/resources/resource_api_token_test.go
vandyliu: we should make this `ReplaceIfConfigured` as well then and mention that in the description

internal/provider/resources/resource_api_token_test.go
```diff
@@ -0,0 +1,716 @@
...
+		Steps: []resource.TestStep{
+			// Test invalid role for token type
+			{
+				Config: astronomerprovider.ProviderConfig(t, true) + apiToken(apiTokenInput{
+					Name: apiTokenName,
+					Type: string(iam.ORGANIZATION),
+					Roles: []apiTokenRole{
+						{
+							Role:       string(iam.WORKSPACEOWNER),
+							EntityId:   workspaceId,
```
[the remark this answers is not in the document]
ichung08: no this is workspaceId intentionally - we want the role to be completely wrong

internal/provider/resources/resource_api_token.go
ichung08: we can't include this validation because `r.OrganizationId` is not set when `ValidateConfig` is run

vandyliu: can we add that as a comment in the code as well

vandyliu (approved):

Issue #112, opened 2024-07-15
Add Team resource


ichung08:
### Description

Add support for team resource on terraform. 

We also need to figure out a strategy for how this will work with scim. Or will not work with scim (disabled when scim is enabled?)

### Acceptance Criteria

The following operations should be supported:
- Create
  - [Create team](https://www.astronomer.io/docs/api/iam-api-reference/team/create-team)
- Update
  - [Update team](https://www.astronomer.io/docs/api/iam-api-reference/team/update-team)
  - [Add team members](https://www.astronomer.io/docs/api/iam-api-reference/team/add-team-members)
  - [Remove team members](https://www.astronomer.io/docs/api/iam-api-reference/team/remove-team-member)
- Delete
  - [Delete team](https://www.astronomer.io/docs/api/iam-api-reference/team/delete-team)

Respective acceptance tests must also be created

PR #118, opened 2024-07-24
Add Team Resource

ichung08:
## Description

Add support for team resource on terraform.

Supports:
- Create
  - [Create team](https://www.astronomer.io/docs/api/iam-api-reference/team/create-team)
- Update
  - [Update team](https://www.astronomer.io/docs/api/iam-api-reference/team/update-team)
  - [Add team members](https://www.astronomer.io/docs/api/iam-api-reference/team/add-team-members)
  - [Remove team members](https://www.astronomer.io/docs/api/iam-api-reference/team/remove-team-member)
- Delete
  - [Delete team](https://www.astronomer.io/docs/api/iam-api-reference/team/delete-team)

## 🎟 Issue(s)

#112 

## 🧪 Functional Testing

Create
Screenshot 2024-07-24 at 3 34 59 PM
Screenshot 2024-07-24 at 3 35 05 PM

Update
Screenshot 2024-07-24 at 4 05 17 PM
Screenshot 2024-07-24 at 4 05 32 PM

Delete
Screenshot 2024-07-24 at 4 06 06 PM

## 📋 Checklist


internal/provider/models/team.go
```diff
@@ -64,3 +80,55 @@ func (data *Team) ReadFromResponse(ctx context.Context, team *iam.Team) diag.Dia
...
+		data.Description = types.StringNull()
+	}
+	if memberIds != nil && len(*memberIds) > 0 {
+		data.MemberIds, diags = utils.StringSet(memberIds)
+		if diags.HasError() {
+			return diags
+		}
+	} else {
+		data.MemberIds = types.SetNull(types.StringType)
+	}
```
ichung08, line 100: wrote it this way to prevent this error:

```
│ Error: Provider produced inconsistent result after apply
│ 
│ When applying changes to astro_team.example1, provider "provider[\"registry.terraform.io/astronomer/astro\"]" produced an unexpected new value: .member_ids: was null, but now cty.SetValEmpty(cty.String).
│ 
│ This is a bug in the provider, which should be reported in the provider's own issue tracker.

```

internal/provider/models/team.go
```diff
@@ -64,3 +80,55 @@ func (data *Team) ReadFromResponse(ctx context.Context, team *iam.Team) diag.Dia
...
+
+func (data *TeamResource) ReadFromResponse(ctx context.Context, team *iam.Team, memberIds *[]string) diag.Diagnostics {
+	var diags diag.Diagnostics
+	data.Id = types.StringValue(team.Id)
+	data.Name = types.StringValue(team.Name)
+	if team.Description != nil && *team.Description != "" {
+		data.Description = types.StringValue(*team.Description)
+	} else {
+		data.Description = types.StringNull()
+	}
```
ichung08, line 92: wrote it this way because i kept hitting `Provider produced inconsistent result after apply`

internal/provider/models/team.go
```diff
@@ -64,3 +80,55 @@ func (data *Team) ReadFromResponse(ctx context.Context, team *iam.Team) diag.Dia
...
+	} else {
+		data.MemberIds = types.SetNull(types.StringType)
+	}
+	data.IsIdpManaged = types.BoolValue(team.IsIdpManaged)
+	orgRole := string(team.OrganizationRole)
+	if orgRole != "" && orgRole != string(iam.ORGANIZATIONMEMBER) {
+		data.OrganizationRole = types.StringValue(orgRole)
+	} else {
+		data.OrganizationRole = types.StringNull()
+	}
```
ichung08, line 107: wrote it this way because i kept hitting `Provider produced inconsistent result after apply` and because the API sets org role to org member automatically if org role is not provided

ichung08:
do we want to support [update team roles](https://www.astronomer.io/docs/api/iam-api-reference/team/update-team-roles) in create and update?

internal/provider/models/team.go
```diff
@@ -26,7 +26,23 @@ type Team struct {
 	UpdatedBy        types.Object `tfsdk:"updated_by"`
 }
 
-func (data *Team) ReadFromResponse(ctx context.Context, team *iam.Team) diag.Diagnostics {
+type TeamResource struct {
+	Id               types.String `tfsdk:"id"`
+	Name             types.String `tfsdk:"name"`
```
vandyliu, line 31: maybe we should also add members info to data source
not in this pr but just a suggestion
we can probably add another rest api call: list team members

internal/provider/models/team.go
```diff
@@ -64,3 +80,55 @@ func (data *Team) ReadFromResponse(ctx context.Context, team *iam.Team) diag.Dia
...
+func (data *TeamResource) ReadFromResponse(ctx context.Context, team *iam.Team, memberIds *[]string) diag.Diagnostics {
+	var diags diag.Diagnostics
+	data.Id = types.StringValue(team.Id)
+	data.Name = types.StringValue(team.Name)
+	if team.Description != nil && *team.Description != "" {
+		data.Description = types.StringValue(*team.Description)
+	} else {
+		data.Description = types.StringNull()
+	}
+	if memberIds != nil && len(*memberIds) > 0 {
```
vandyliu, line 93: prob dont need this if statement since the util function covers it

internal/provider/models/team.go
```diff
@@ -64,3 +80,55 @@ func (data *Team) ReadFromResponse(ctx context.Context, team *iam.Team) diag.Dia
...
+		data.MemberIds, diags = utils.StringSet(memberIds)
+		if diags.HasError() {
+			return diags
+		}
+	} else {
+		data.MemberIds = types.SetNull(types.StringType)
+	}
+	data.IsIdpManaged = types.BoolValue(team.IsIdpManaged)
```
vandyliu, line 101: shouldn't this always be false

vandyliu: wait nvm

internal/provider/schemas/team.go
```diff
@@ -67,3 +74,81 @@ func TeamDataSourceSchemaAttributes() map[string]datasourceSchema.Attribute {
...
+				Attributes: ResourceWorkspaceRoleSchemaAttributes(),
+			},
+			Computed:            true,
+			MarkdownDescription: "The roles assigned to the workspaces",
+		},
+		"deployment_roles": resourceSchema.SetNestedAttribute{
+			NestedObject: resourceSchema.NestedAttributeObject{
+				Attributes: ResourceDeploymentRoleSchemaAttributes(),
+			},
+			Computed:            true,
```
vandyliu, line 128: should prob be optional or maybe even required but can be empty

internal/provider/schemas/team.go
```diff
@@ -67,3 +74,81 @@ func TeamDataSourceSchemaAttributes() map[string]datasourceSchema.Attribute {
...
+		"description": resourceSchema.StringAttribute{
+			MarkdownDescription: "Team description",
+			Optional:            true,
+		},
+		"member_ids": resourceSchema.SetAttribute{
+			ElementType:         types.StringType,
+			MarkdownDescription: "The IDs of the users to add to the Team",
+			Optional:            true,
+			Validators: []validator.Set{
+				setvalidator.SizeAtLeast(1),
```
vandyliu, line 100: does it have to be 1, can we have empty team?

internal/provider/schemas/team.go
```diff
@@ -67,3 +74,81 @@ func TeamDataSourceSchemaAttributes() map[string]datasourceSchema.Attribute {
...
+		},
+		"description": resourceSchema.StringAttribute{
+			MarkdownDescription: "Team description",
+			Optional:            true,
+		},
+		"member_ids": resourceSchema.SetAttribute{
+			ElementType:         types.StringType,
+			MarkdownDescription: "The IDs of the users to add to the Team",
+			Optional:            true,
+			Validators: []validator.Set{
```
vandyliu, line 99: lets also make sure that they are cuids

internal/provider/models/team.go
ichung08: the issue i ran into was that the util function returns `types.SetValue(types.StringType, []attr.Value{})` for nil values

but what i need is a null set, which is `types.SetNull(types.StringType)`

ichung08: i'll keep testing with and without the guard but i ran into issues when i didn't do it this way

internal/provider/schemas/team.go
ichung08: yes we can have empty team - will change

internal/provider/models/team.go
vandyliu: makes sense
can leave it as is if it errors

internal/provider/schemas/team.go
ichung08: changed to optional - should not be required as teams don't have to have `deployment_roles`

internal/provider/common/role.go
```diff
@@ -0,0 +1,70 @@
+package common
```
ichung08, line 1: added this file for shared role functions

needed to create a new directory because if i had it in utils i would hit an `import cycle` error since models, resources and utils reference each other in a cycle

internal/provider/provider_test_utils.go
```diff
@@ -47,11 +47,14 @@ func TestAccPreCheck(t *testing.T) {
 	}
 }
 
-func ProviderConfig(t *testing.T, isHosted bool) string {
+func ProviderConfig(t *testing.T, isHosted bool, isScim bool) string {
```
ichung08, line 50: added this param to the signature and had to update all the files accordingly (which is why all the test files were affected)

examples/resources/astro_team/resource.tf
```diff
@@ -0,0 +1,20 @@
+resource "astro_team" "example" {
+  name              = "team"
+  description       = "team-description"
+  member_ids        = ["clhpichn8002m01mqa4ocs7g6"]
+  organization_role = "ORGANIZATION_OWNER"
+  workspace_role = [{
+    workspace_id = "clx42sxw501gl01o0gjenthnh"
+    role         = "WORKSPACE_OWNER"
+  }]
+  deployment_role = [{
```
vandyliu, line 10:
```suggestion
  deployment_roles = [{
```

internal/provider/provider_test_utils.go
```diff
@@ -47,11 +49,14 @@ func TestAccPreCheck(t *testing.T) {
...
 	var orgId, token string
 	if isHosted {
 		orgId = os.Getenv("HOSTED_ORGANIZATION_ID")
 		token = os.Getenv("HOSTED_ORGANIZATION_API_TOKEN")
+	} else if isScim {
```
vandyliu, line 57: actually it may be best to make this an enum that takes in `HOSTED` `SCIM` or `HYBRID`
so something like `ProviderConfig(t *testing.T, testOrganizationType string)` or `ProviderConfig(t *testing.T, testOrganizationType TestOrganizationType)`

internal/provider/resources/resource_team.go
```diff
@@ -0,0 +1,602 @@
...
+		return diags
+	}
+	_, diagnostic := clients.NormalizeAPIError(ctx, listDeployments.HTTPResponse, listDeployments.Body)
+	if diagnostic != nil {
+		diags.Append(diagnostic)
+		return diags
+	}
+
+	// get list of workspace ids from deployments
+	deploymentWorkspaceIds := lo.Map(listDeployments.JSON200.Deployments, func(deployment platform.Deployment, _ int) string {
```
vandyliu, line 123: we should make sure `JSON200 != nil`

internal/provider/resources/resource_team.go
```diff
@@ -0,0 +1,602 @@
...
+				"Not every deployment role has a corresponding workspace role. Missing workspace roles for deploymentWorkspaceIds: %v", missingWorkspaceIds),
+		})
+		diags.AddError(
+			"Bad Request Error",
+			fmt.Sprintf("Unable to mutate Team roles, not every deployment role has a corresponding workspace role. Add workspace roles for the missing workspace ids: %v", missingWorkspaceIds),
+		)
+		return diags
+	}
+
+	// create request
```
vandyliu, line 146: outdated comment

internal/provider/resources/resource_team.go
```diff
@@ -0,0 +1,602 @@
...
+	}
+
+	// Create the team request
+	createTeamRequest := iam.CreateTeamRequest{
+		Name:        data.Name.ValueString(),
+		Description: data.Description.ValueStringPointer(),
+		MemberIds:   &memberIds,
+	}
+
+	if !data.OrganizationRole.IsNull() {
```
vandyliu, line 204: shuold never be null so prob dont need this and can add it to the initialization

internal/provider/resources/resource_team.go
```diff
@@ -0,0 +1,602 @@
...
+	resp *resource.ValidateConfigResponse,
+) {
+	var data models.TeamResource
+
+	resp.Diagnostics.Append(req.Config.Get(ctx, &data)...)
+	if resp.Diagnostics.HasError() {
+		return
+	}
+
+	// Validate if org isScimEnabled
```
vandyliu, line 551:
```suggestion
	// Validate if org isScimEnabled and error if it is
```

internal/provider/resources/resource_team.go
```diff
@@ -0,0 +1,602 @@
...
+	}
+
+	for _, role := range deploymentRoles {
+		if !common.ValidateRoleMatchesEntityType(role.Role, string(iam.DEPLOYMENT)) {
+			resp.Diagnostics.AddError(
+				fmt.Sprintf("Role '%s' is not valid for role type '%s'", role.Role, string(iam.DEPLOYMENT)),
+				fmt.Sprintf("Please provide a valid role for the type '%s'", string(iam.DEPLOYMENT)),
+			)
+			return
+		}
```
vandyliu, line 600: do we have a check where they cannot use same workspaceId or deploymentId for the role
eg. 
workspace_roles = [{workspace_id: cuid1, role: workspace_member}, {workspace_id: cuid1, role: workspace_owner}]

internal/provider/resources/resource_team_roles.go
```diff
@@ -82,15 +87,60 @@ func (r *teamRolesResource) MutateRoles(
...
+	// check if deploymentWorkspaceIds are in workspaceIds
+	workspaceIds = lo.Intersect(lo.Uniq(workspaceIds), lo.Uniq(deploymentWorkspaceIds))
+	if len(workspaceIds) != len(deploymentWorkspaceIds) {
+		tflog.Error(ctx, "failed to mutate Team roles", map[string]interface{}{"error": err})
+		diags.AddError(
+			"Client Error",
+			fmt.Sprintf("Unable to mutate Team roles, not every deployment role has a corresponding workspace role"),
+		)
+		return diags
+	}
```
vandyliu, line 142: can we make this into a helper so we can reuse in `resource_team_roels` and `resource_team`

internal/provider/resources/resource_team_test.go
```diff
@@ -0,0 +1,237 @@
...
+	resourceVar := fmt.Sprintf("astro_team.%v", teamName)
+
+	resource.Test(t, resource.TestCase{
+		ProtoV6ProviderFactories: astronomerprovider.TestAccProtoV6ProviderFactories,
+		PreCheck:                 func() { astronomerprovider.TestAccPreCheck(t) },
+		CheckDestroy: resource.ComposeTestCheckFunc(
+			testAccCheckTeamExistence(t, teamName, false),
+		),
+		Steps: []resource.TestStep{
+			// Test failure: disable team resource if org is isScimEnabled
```
vandyliu, line 38: lets add test where we have deployment but no matching workspace and make sure we get the right err

internal/provider/resources/resource_team.go
ichung08: no but i can add one

.github/workflows/testacc.yml
```diff
@@ -75,21 +75,23 @@ jobs:
              break
            fi
          done
-         echo "SKIP_CLUSTER_RESOURCE_TESTS=$SKIP_TESTS"
-         echo "SKIP_CLUSTER_RESOURCE_TESTS=$SKIP_TESTS" >> $GITHUB_ENV
+         echo "SKIP_CLUSTER_RESOURCE_TESTS=True"
+         echo "SKIP_CLUSTER_RESOURCE_TESTS=True" >> $GITHUB_ENV
```
ichung08, line 79: TODO: revert

internal/provider/models/team.go
```diff
@@ -10,12 +10,13 @@ import (
...
 
-// Team describes the data source data model.
-type Team struct {
+// TeamDataSource describes the data source data model.
+type TeamDataSource struct {
 	Id               types.String `tfsdk:"id"`
 	Name             types.String `tfsdk:"name"`
 	Description      types.String `tfsdk:"description"`
 	IsIdpManaged     types.Bool   `tfsdk:"is_idp_managed"`
+	OrganizationId   types.String `tfsdk:"organization_id"`
```
ichung08, line 19: discovered we were actually missing `OrganizationId` and added this field here and to schema and tests

internal/provider/schemas/team.go
```diff
@@ -67,3 +78,85 @@ func TeamDataSourceSchemaAttributes() map[string]datasourceSchema.Attribute {
...
+				setvalidator.ValueStringsAre(validators.IsCuid()),
+			},
+		},
+		"is_idp_managed": resourceSchema.BoolAttribute{
+			MarkdownDescription: "Whether the Team is managed by an identity provider",
+			Computed:            true,
+		},
+		"organization_id": resourceSchema.StringAttribute{
```
vandyliu, line 111: i don't think we need this
we should use r.OrganizationId in `Create` and `Update` - they should exist there at runtime

internal/provider/schemas/teams.go
```diff
@@ -15,6 +15,7 @@ func TeamsElementAttributeTypes() map[string]attr.Type {
 		"name":              types.StringType,
 		"description":       types.StringType,
 		"is_idp_managed":    types.BoolType,
+		"organization_id":   types.StringType,
```
vandyliu, line 18: prob dont need this

internal/provider/schemas/team.go
```diff
@@ -22,29 +29,33 @@ func TeamDataSourceSchemaAttributes() map[string]datasourceSchema.Attribute {
 			Computed:            true,
 		},
 		"is_idp_managed": datasourceSchema.BoolAttribute{
-			MarkdownDescription: "Whether the team is managed by an identity provider",
+			MarkdownDescription: "Whether the Team is managed by an identity provider",
+			Computed:            true,
+		},
+		"organization_id": datasourceSchema.StringAttribute{
+			MarkdownDescription: "The ID of the Organization to which the Team belongs",
```
vandyliu, line 36: shouldn't need this

internal/provider/models/team.go
vandyliu: i purposely removed it since its clear that the organizationId will be the same one that is configured in the provider config

internal/provider/common/role.go
```diff
@@ -0,0 +1,170 @@
...
+
+type ValidateWorkspaceDeploymentRolesInput struct {
+	PlatformClient  *platform.ClientWithResponses
+	OrganizationId  string
+	DeploymentRoles []iam.DeploymentRole
+	WorkspaceRoles  []iam.WorkspaceRole
+}
+
+// ValidateWorkspaceDeploymentRoles checks if deployment roles have corresponding workspace roles
+func ValidateWorkspaceDeploymentRoles(ctx context.Context, diags diag.Diagnostics, input ValidateWorkspaceDeploymentRolesInput) diag.Diagnostics {
```
vandyliu, line 86:
```suggestion
func ValidateWorkspaceDeploymentRoles(ctx context.Context, input ValidateWorkspaceDeploymentRolesInput) diag.Diagnostics {
```

vandyliu: can make a new variable instead of passing it in as arg

internal/provider/common/role.go
```diff
@@ -0,0 +1,170 @@
...
+			duplicates = append(duplicates, id)
+		}
+	}
+
+	return duplicates
+}
+
+// CheckDuplicateDeploymentId checks if there are duplicate deployment ids in the deployment roles
+func CheckDuplicateDeploymentId(deploymentRoles []iam.DeploymentRole) []string {
+	workspaceIdCount := make(map[string]int)
```
vandyliu, line 157:
```suggestion
	deploymentIdCount := make(map[string]int)
```

internal/provider/common/role.go
```diff
@@ -0,0 +1,170 @@
...
+	for _, role := range workspaceRoles {
+		workspaceIdCount[role.WorkspaceId]++
+	}
+
+	var duplicates []string
+	for id, count := range workspaceIdCount {
+		if count > 1 {
+			duplicates = append(duplicates, id)
+		}
+	}
```
vandyliu, line 150: we can just do lo.UniqBy and then make sure the lengths are equal

internal/provider/resources/resource_team_roles.go
```diff
@@ -82,11 +87,22 @@ func (r *teamRolesResource) MutateRoles(
...
 		return diags
 	}
-	deploymentRoles, diags := RequestDeploymentRoles(ctx, data.DeploymentRoles)
+	deploymentRoles, diags := common.RequestDeploymentRoles(ctx, data.DeploymentRoles)
+	if diags.HasError() {
+		return diags
+	}
+
+	// Validate the roles
+	diags = common.ValidateWorkspaceDeploymentRoles(ctx, diags, common.ValidateWorkspaceDeploymentRolesInput{
```
vandyliu, line 100:
```suggestion
	diags = common.ValidateWorkspaceDeploymentRoles(ctx, common.ValidateWorkspaceDeploymentRolesInput{
```

internal/provider/resources/resource_team.go
```diff
@@ -0,0 +1,619 @@
...
+	r.PlatformClient = apiClients.PlatformClient
+	r.OrganizationId = apiClients.OrganizationId
+}
+
+func (r *TeamResource) MutateRoles(
+	ctx context.Context,
+	data *models.TeamResource,
+	teamId string,
+) diag.Diagnostics {
+	// Then convert the models to the request types for the API
```
vandyliu, line 87:
```suggestion
	// convert the models to the request types for the API
```

internal/provider/resources/resource_team.go
```diff
@@ -0,0 +1,619 @@
...
+	if diagnostic != nil {
+		resp.Diagnostics.Append(diagnostic)
+		return
+	}
+
+	memberIds, diags := utils.TypesSetToStringSlice(ctx, data.MemberIds)
+	if diags.HasError() {
+		resp.Diagnostics.Append(diags...)
+		return
+	}
```
vandyliu, line 276: shouldn't this be done in `data.ReadFromResponse`

vandyliu: nvm i see why

internal/provider/models/team.go
ichung08: ah... ok

internal/provider/common/role.go
```diff
@@ -0,0 +1,170 @@
...
+			"Unable to mutate Team roles, not every deployment role has a corresponding workspace role",
+			"Please ensure that every deployment role has a corresponding workspace role",
+		)
+		return diags
+	}
+	return diags
+}
+
+// CheckDuplicateWorkspaceId checks if there are duplicate workspace ids in the workspace roles
+func CheckDuplicateWorkspaceId(workspaceRoles []iam.WorkspaceRole) []string {
```
vandyliu, line 139:
```suggestion
func HasDuplicateWorkspaceIds(workspaceRoles []iam.WorkspaceRole) bool {
```

internal/provider/common/role.go
```diff
@@ -0,0 +1,170 @@
...
+		if count > 1 {
+			duplicates = append(duplicates, id)
+		}
+	}
+
+	return duplicates
+}
+
+// CheckDuplicateDeploymentId checks if there are duplicate deployment ids in the deployment roles
+func CheckDuplicateDeploymentId(deploymentRoles []iam.DeploymentRole) []string {
```
vandyliu, line 156:
```suggestion
func HasDuplicateDeploymentId(deploymentRoles []iam.DeploymentRole) bool {
```

internal/provider/common/role.go
ichung08: i did it this way so we can return what the duplicate ids are in the error message

internal/provider/common/role.go
```diff
@@ -0,0 +1,170 @@
...
+
+// ValidateRoleMatchesEntityType checks if the role is valid for the entityType
+func ValidateRoleMatchesEntityType(role string, scopeType string) bool {
+	organizationRoles := []string{string(iam.ORGANIZATIONBILLINGADMIN), string(iam.ORGANIZATIONMEMBER), string(iam.ORGANIZATIONOWNER)}
+	workspaceRoles := []string{string(iam.WORKSPACEACCESSOR), string(iam.WORKSPACEAUTHOR), string(iam.WORKSPACEMEMBER), string(iam.WORKSPACEOWNER), string(iam.WORKSPACEOPERATOR)}
+	var roles []string
+
+	scopeType = strings.ToLower(scopeType)
+	if scopeType == "organization" {
+		roles = organizationRoles
```
aliotta, line 67: can we make these checks similar to how we handle deployment (e.g. nonWorkspaceDeploymentRoles) as we will eventually make custom org roles and this will break when that happens.

internal/provider/common/role.go
```diff
@@ -0,0 +1,170 @@
...
+	} else if scopeType == "workspace" {
+		roles = workspaceRoles
```
aliotta, line 69: same as above comment

internal/provider/resources/resource_team_test.go
```diff
@@ -0,0 +1,284 @@
...
+						},
+					},
+					WorkspaceRoles: []role{
+						{
+							Role: string(iam.WORKSPACEOWNER),
+							Id:   workspaceId,
+						},
+					},
+				}),
+				ExpectError: regexp.MustCompile("Invalid Configuration: Cannot create, update or delete a Team resource when SCIM is enabled"),
```
aliotta, line 60: Thank you!

internal/provider/common/role.go
```diff
@@ -126,17 +126,17 @@ func ValidateWorkspaceDeploymentRoles(ctx context.Context, diags diag.Diagnostic
...
+		}
 	}
-	return diags
+	return nil
 }
 
-// CheckDuplicateWorkspaceId checks if there are duplicate workspace ids in the workspace roles
-func CheckDuplicateWorkspaceId(workspaceRoles []iam.WorkspaceRole) []string {
+// HasDuplicateWorkspaceId checks if there are duplicate workspace ids in the workspace roles
+func HasDuplicateWorkspaceId(workspaceRoles []iam.WorkspaceRole) []string {
```
vandyliu, line 139: since we're actually still returning the list
```suggestion
func GetDuplicateWorkspaceIds(workspaceRoles []iam.WorkspaceRole) []string {
```

internal/provider/common/role.go
```diff
@@ -152,15 +152,15 @@ func CheckDuplicateWorkspaceId(workspaceRoles []iam.WorkspaceRole) []string {
 	return duplicates
 }
 
-// CheckDuplicateDeploymentId checks if there are duplicate deployment ids in the deployment roles
-func CheckDuplicateDeploymentId(deploymentRoles []iam.DeploymentRole) []string {
-	workspaceIdCount := make(map[string]int)
+// HasDuplicateDeploymentId checks if there are duplicate deployment ids in the deployment roles
+func HasDuplicateDeploymentId(deploymentRoles []iam.DeploymentRole) []string {
```
vandyliu, line 156:
```suggestion
func GetDuplicateDeploymentIds(deploymentRoles []iam.DeploymentRole) []string {
```

vandyliu (approved):
one last comment but looks good

Issue #114, opened 2024-07-15
Add User Roles Resource


ichung08:
### Description

Add support for user roles on terraform.

### Acceptance Criteria

The following operations should be supported:
- Update
  - [Update user roles](https://www.astronomer.io/docs/api/iam-api-reference/user/update-user-roles)

Respective acceptance tests must also be created 

### Anything else?

Reference [`team_roles`](https://github.com/astronomer/terraform-provider-astro/pull/69)

PR #119, opened 2024-08-01
Add User Roles Resource

ichung08:
## Description

Add support on terraform for user roles resource.

The following operations should be supported:

- Update
  - [Update user roles](https://www.astronomer.io/docs/api/iam-api-reference/user/update-user-roles)

## 🎟 Issue(s)

#114 

## 🧪 Functional Testing

Screenshot 2024-07-31 at 3 50 41 PM

## 📋 Checklist


.github/workflows/testacc.yml
```diff
@@ -92,6 +92,7 @@ jobs:
           SKIP_CLUSTER_RESOURCE_TESTS: ${{ env.SKIP_CLUSTER_RESOURCE_TESTS }}
           HOSTED_TEAM_ID: clx44rvzr01nc01o06pze6qb7
           HOSTED_USER_ID: clz3a4ymt004x01on8w5ydq8j
+          HOSTED_DUMMY_USER_ID: clzawipbm00bm01qw98vzzoca
```
ichung08, line 95: added new user id with no team or roles - using existing user causes conflicts with tests since they've already been assigned multiple roles/teams

vandyliu: can we make it obvious that this user is `HOSTED_NON_TEAMS_USER_ID` and the other one is `HOSTED_HAS_TEAMS_USER_ID` or something like that

internal/provider/common/role.go
```diff
@@ -16,6 +16,11 @@ import (
 	"github.com/samber/lo"
 )
 
+type Role struct {
+	Role string
+	Id   string
```
vandyliu, line 21:
```suggestion
	EntityId   string
```

internal/provider/common/role.go
```diff
@@ -172,3 +177,45 @@ func GetDuplicateDeploymentIds(deploymentRoles []iam.DeploymentRole) []string {
...
+// ContainsDeploymentRoles checks if a list of deployment roles contains a list of roles
+func ContainsDeploymentRoles(userRoles []iam.DeploymentRole, roles []Role) []Role {
+	missingRoles := []Role{}
+	for _, role := range roles {
+		if !ContainsDeploymentRole(userRoles, role) {
+			missingRoles = append(missingRoles, role)
+		}
+	}
+	return missingRoles
+}
```
vandyliu, line 221: seems that these are only used in test files
can we move it to a directory just for tests

internal/provider/common/role.go
```diff
@@ -16,6 +16,11 @@ import (
 	"github.com/samber/lo"
 )
 
+type Role struct {
+	Role string
+	Id   string
+}
```
vandyliu, line 22: we should also move this to a directory just for tests
don't want it to accidentally be used in the business logic since Role and EntityId could be used incorrectly if we don't know the entityType

vandyliu (approved):

Issue #121, opened 2024-08-02
Add a new scheduler size `EXTRA_LARGE` to terraform deployment schema


ichung08:
### Description

add a new scheduler size “EXTRA_LARGE” to terraform deployment schema
change is in [deployment.go](https://github.com/astronomer/terraform-provider-astro/blob/1697d2e0bb7bc36c4084513a7ce925d2c9e08346/internal/provider/schemas/deployment.go#L242) schema file

requires update to terraform core api spec

[slack thread](https://astronomer.slack.com/archives/C07CHFRM9B2/p1722434859920319)

PR #122, opened 2024-08-03
Add a new scheduler size to deployment schema

ichung08:
## Description

Add a new scheduler size to deployment schema.
Also update terraform's core API spec and corresponding models, schemas and resources.

## 🎟 Issue(s)

#121 

## 📋 Checklist


internal/provider/schemas/deployment.go
```diff
@@ -247,6 +247,7 @@ func DeploymentResourceSchemaAttributes() map[string]resourceSchema.Attribute {
 					string(platform.SchedulerMachineNameSMALL),
 					string(platform.SchedulerMachineNameMEDIUM),
 					string(platform.SchedulerMachineNameLARGE),
+					string(platform.SchedulerMachineNameEXTRALARGE),
```
vandyliu, line 250: we should have a test that uses this value

vandyliu (approved):

PR #142, opened 2024-08-23
Fix Bug in Role Helper

ichung08:
## Description

Fixes bug in `ValidateWorkspaceDeploymentRoles` where users weren't able to assign multiple deployment roles under the same workspace

## 📋 Checklist

### DIFF
--- a/internal/provider/common/role.go
+++ b/internal/provider/common/role.go
@@ -120,15 +120,15 @@ func ValidateWorkspaceDeploymentRoles(ctx context.Context, input ValidateWorkspa
 	deploymentWorkspaceIds := lo.Map(listDeployments.JSON200.Deployments, func(deployment platform.Deployment, _ int) string {
 		return deployment.WorkspaceId
 	})
-	//deploymentWorkspaceIds = lo.Uniq(deploymentWorkspaceIds)
+	deploymentWorkspaceIds = lo.Uniq(deploymentWorkspaceIds)
+
 	// get list of workspaceIds
 	workspaceIds := lo.Map(input.WorkspaceRoles, func(role iam.WorkspaceRole, _ int) string {
 		return role.WorkspaceId
 	})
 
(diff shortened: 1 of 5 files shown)


internal/provider/common/role.go
```diff
@@ -120,14 +120,15 @@ func ValidateWorkspaceDeploymentRoles(ctx context.Context, input ValidateWorkspa
...
+	deploymentWorkspaceIds = lo.Uniq(deploymentWorkspaceIds)
 
 	// get list of workspaceIds
 	workspaceIds := lo.Map(input.WorkspaceRoles, func(role iam.WorkspaceRole, _ int) string {
 		return role.WorkspaceId
 	})
 
 	// check if deploymentWorkspaceIds are in workspaceIds
-	workspaceIds = lo.Intersect(lo.Uniq(workspaceIds), lo.Uniq(deploymentWorkspaceIds))
+	workspaceIds = lo.Intersect(lo.Uniq(workspaceIds), deploymentWorkspaceIds)
```
vandyliu, line 131: can we have a test case where before this change it would fail and after it would pass

internal/provider/resources/resource_user_roles_test.go
```diff
@@ -55,6 +55,30 @@ func TestAcc_ResourceUserRoles(t *testing.T) {
...
+							},
+						},
+						DeploymentRoles: []utils.Role{
+							{
+								Role:     "DEPLOYMENT_ADMIN",
+								EntityId: deploymentId,
+							},
+							{
+								Role:     "DEPLOYMENT_ADMIN",
+								EntityId: "cm070pg0r00wd01qgnskk0dir",
```
vandyliu, line 76: where did this `deploymentId` come from?
should we create a deployment for this test and use that one instead?

vandyliu (approved):

Issue #189, opened 2025-01-12
Allow setting workload identity with "astro_deployment"


yfried:
### Description

Please allow astro_deployment resource to set the deployment workload-identity, instead of forcing user to hack something with cli/api after the deployment is created

### Acceptance Criteria

`workload_identity` attribute of `astro_deployment` becomes a setter, instead of just a read-only

chobberoni:
Resolved via https://github.com/astronomer/terraform-provider-astro/pull/206

PR #206, opened 2025-02-06
add workload identity support

aliotta:
## Description

Adds support for workload identity via a option desired workload identity param.

[#23877](https://github.com/astronomer/astro/issues/23877)

## 📸 Screenshots

On create no errors:
Screenshot 2025-02-06 at 9 48 58 AM
On update of a non-workload identity param no errors:
Screenshot 2025-02-06 at 9 49 03 AM
On update of a workload identity param no errors:
Screenshot 2025-02-06 at 9 49 03 AM
UI on create:
Screenshot 2025-02-06 at 9 49 31 AM
UI on update:
Screenshot 2025-02-06 at 9 52 24 AM
Generated .tf from import script: 
Screenshot 2025-02-06 at 2 58 40 PM
Updating the workload identity: 
Screenshot 2025-02-06 at 2 58 57 PM
Results of update in ui:
Screenshot 2025-02-06 at 2 59 16 PM

## 📋 Checklist


aliotta:
> change looks good overall, can you also test importing a deployment without workload identity and updating that?

good call I did need to make changes to the import script. One thing to note though is that deployments will always have a workload identity as if one is not sent the backend makes one for you.

internal/provider/schemas/deployment.go
```diff
@@ -195,6 +195,13 @@ func DeploymentResourceSchemaAttributes() map[string]resourceSchema.Attribute {
 				stringplanmodifier.UseStateForUnknown(),
 			},
 		},
+		"desired_workload_identity": resourceSchema.StringAttribute{
+			MarkdownDescription: "Deployment's desired workload identity. The Terraform provider will use this provided workload identity to create the Deployment. If it is not provided the worload identity will be assigned automatically.",
```
vandyliu, line 199:
```suggestion
			MarkdownDescription: "Deployment's desired workload identity. The Terraform provider will use this provided workload identity to create the Deployment. If it is not provided the workload identity will be assigned automatically.",
```

internal/provider/schemas/deployment.go
```diff
@@ -195,6 +195,13 @@ func DeploymentResourceSchemaAttributes() map[string]resourceSchema.Attribute {
...
+			Optional:            true,
+			PlanModifiers: []planmodifier.String{
+				stringplanmodifier.RequiresReplaceIfConfigured(),
```
vandyliu, line 202: is this needed?

aliotta: not sure i'll test it without to see.

aliotta: good eyes. We do not want this as it causes the deployment to get deleted and recreated on change.

aliotta: I retested the update flow after removing this and all looks good.

vandyliu (approved):

yfried:
thank you for this much needed feature.
Small question: why `desired_workload_identity` instead of simply `workload_identity`?

chobberoni:
closes https://github.com/astronomer/terraform-provider-astro/issues/189

PR #227, opened 2025-06-30
regen core platform client from astro @ 688782a

jaketf:
## Description

regenerate the client for core platform client in prep for the programatic support epic so it supports the Astro Executor and Remote Execution blocks.
this PR just regenerates the clients and makes a pointer change so `go vet` is happy. no logic changes. 

## 🎟 Issue(s)
None.

## 📋 Checklist
None of these really need to be updated as this is essentially an autogenerated PR.
When we actually use the new fields we'll add more tests and screenshots verifying the functionality.


feluelle (approved):

ianbuss:
@jaketf this is just the internal client?

Issue #220, opened 2025-05-23
add remote execution support


jaketf:
### Description

add support for the remote execution block 

```
  // for airflow 3 and above an optional section for remote execution
  // - if missing on create, remote execution will be disabled
  // - if null on update, remote execution configuration will be unchanged
  // - if provided, all sub-fields are set to the values provided in the
  //   update (following the existing convention on scalingSpec)
  // to disable remote execution, the remoteExecution.enabled field
  // must be set to false on an update
  "remoteExecution": {          // optional section
    // if remote execution is enabled for the deployment
    "enabled": true|false,      // required
    
    // the allowed source IP ranges for remote agent traffic
    "allowedIpAddressRanges": [     // optional, default to no access
	    "1.2.3.4",
	    "2.3.4.5/32" // accept both CIDR and single IPs
	  ],
	  
	  // remote task log configuration
	  // if unset, the hosted components cannot serve logs directly
	  // if set, workloadIdentity should also be set
	  // mutually exclusive with taskLogUrlTemplate
	  "taskLogBucket": "s3://..."     // optional, 
	  
	  // in the case where logs are stored remotely the user can provide 
	  // a template URL to their custom location to allow a custom Astro
	  // UI plugin to display a URL link for the task logs. Should accept
	  // the same templating parameters as the Airflow filename template
	  // settings
	  "taskLogUrlTemplate": ""        // optional
  }
```

depends on https://github.com/astronomer/astro/issues/30901

### Acceptance Criteria

- [] can create and update deployments with remote execution details

PR #228, opened 2025-07-07
add astro executor support

jaketf:
## Description

add support for astro executor for standard hosted and dedicated deployments.

## 🎟 Issue(s)

closes: #219 

## 📋 Checklist

- [x] Added/updated applicable tests
- [x] Added/updated examples in the `examples/` directory
- [ ] Updated any related [documentation](https://github.com/astronomer/docs/)

PR #231, opened 2025-07-09
Add remote execution support

feluelle:
## Description

depends on: https://github.com/astronomer/terraform-provider-astro/pull/228

Add remote execution support to `astro_deployment` resource and data objects.

## 🎟 Issue(s)

closes: https://github.com/astronomer/terraform-provider-astro/issues/220

## 📋 Checklist


internal/provider/resources/resource_deployment.go
```diff
@@ -1042,6 +1068,41 @@ func RequestScalingSpec(ctx context.Context, scalingSpecObj types.Object) (*plat
 	return platformScalingSpec, nil
 }
 
+// RequestRemoteExecution converts a Terraform object to a platform.RemoteExecutionRequest to be used in create and update requests
+func RequestRemoteExecution(ctx context.Context, remoteExecutionObj types.Object) (*platform.DeploymentRemoteExecutionRequest, diag.Diagnostics) {
```
ichung08, line 1072: could you also add tests for this to the `resource_deployment_test` file? generic deployment tests and some tests on input validation would also be good (eg. validating that remote execution is only allowed on dedicated)

internal/provider/resources/resource_deployment.go
```diff
@@ -850,25 +876,25 @@ func validateHostedConfig(ctx context.Context, data *models.DeploymentResource)
 			"Please provide is_development_mode",
 		)
 	}
-	if data.ResourceQuotaCpu.IsNull() {
+	if data.ResourceQuotaCpu.IsNull() && data.RemoteExecution.IsNull() {
```
ichung08, line 879: why do we need all these additional remote execution null checks?

ichung08 (commented):
overall looks good, just one question

internal/provider/resources/resource_deployment.go
feluelle: Because they cannot be set for Remote Execution Deployments.

internal/provider/models/deployment.go
```diff
@@ -577,3 +587,32 @@ func ScalingSpecTypesObject(
...
+		return types.ObjectNull(schemas.RemoteExecutionAttributeTypes()), nil
+	}
+	allowedIpAddressRanges, diags := utils.StringSet(&remoteExecution.AllowedIpAddressRanges)
+	if diags.HasError() {
+		return types.ObjectNull(schemas.RemoteExecutionAttributeTypes()), diags
+	}
+	obj := RemoteExecution{
+		Enabled:                types.BoolValue(remoteExecution.Enabled),
+		AllowedIpAddressRanges: allowedIpAddressRanges,
+		RemoteApiUrl:           types.StringValue(remoteExecution.RemoteApiUrl),
```
neel-astro, line 613: just trying to learn the ropes: Is this request struct being generated, or API response being converted to an object for terraform SDK to handle?
For requests, users don't need to pass remote API url

neel-astro (approved):
Not a Terraform SDK expert, but this has been a good learning PR for me. I have only validated that the remote execution config is in line with the API contract.

internal/provider/models/deployment.go
ichung08: this is the latter, the API response is being converted into an object for terraform to handle and read

Issue #254, opened 2025-10-06
[Bug]:Terraform provider failing to handle unknown status from cluster


sakethsomaraju:
### Description

The Terraform provider fails to handle the status from the provider.

> When initiating the cluster creation using the Terraform, the astro_cluster resource fails immediately with an error “unexpected cluster status '' for cluster”. Although the cluster appears in Astro and is created correctly with the proper range, Terraform does not wait for the cluster to be running and subsequently breaks.

<img width="839" height="549" alt="Image" src="https://github.com/user-attachments/assets/cc6811c0-abcd-4987-a3f9-0834e7a42833" />

<img width="839" height="206" alt="Image" src="https://github.com/user-attachments/assets/97bb74ce-d0da-4959-b96b-f359ea0ae54b" />

### Steps To Reproduce

Add the below resource.:

```

resource "astro_cluster" "azure_example" {
  type             = "DEDICATED"
  name             = "my first azure cluster"
  region           = "westus2"
  cloud_provider   = "AZURE"
  vpc_subnet_range = "172.20.0.0/19"
  workspace_ids    = []
}
```

Perform terraform plan:

```
terraform plan

Terraform used the selected providers to generate the following execution plan. Resource actions are indicated with the following symbols:
  + create

Terraform will perform the following actions:

  # astro_cluster.azure_example will be created
  + resource "astro_cluster" "azure_example" {
      + cloud_provider   = "AZURE"
      + created_at       = (known after apply)
      + db_instance_type = (known after apply)
      + health_status    = (known after apply)
      + id               = (known after apply)
      + is_limited       = (known after apply)
      + metadata         = (known after apply)
      + name             = "my first azure cluster"
      + node_pools       = (known after apply)
      + provider_account = (known after apply)
      + region           = "westus2"
      + status           = (known after apply)
      + tenant_id        = (known after apply)
      + type             = "DEDICATED"
      + updated_at       = (known after apply)
      + vpc_subnet_range = "172.20.0.0/19"
      + workspace_ids    = []
    }

Plan: 1 to add, 0 to change, 0 to destroy.

```
Run Terraform apply:
```
terraform apply

Terraform used the selected providers to generate the following execution plan. Resource actions are indicated with the following symbols:
  + create

Terraform will perform the following actions:

  # astro_cluster.azure_example will be created
  + resource "astro_cluster" "azure_example" {
      + cloud_provider   = "AZURE"
      + created_at       = (known after apply)
[...]

PR #255, opened 2025-10-10
Fix issue where cluster creation fails immediately due to empty status response

cjonesy:
## Description

It seems that the API is returning an empty string for status for a very short period of time when the cluster is first created.

This feels like a bug in the API, however it can be addressed on the provider side if we continue to poll for status changes when we get a status of `""`.

## 🎟 Issue(s)

#254

## 🧪 Functional Testing

I have tested these changes in my environment and the cluster was successfully created.

I also added tests for all the various statuses in common_cluster.

## 📸 Screenshots

From the linked issue:
image
image

From my testing of the updated provider:
image
image

## 📋 Checklist


cjonesy:
I tested again this morning and it seems the underlying issue with the API has been fixed. This PR is no longer needed!

Issue #244, opened 2025-09-08
[Bug]: Inconsistent result after apply for `remote_execution.allowed_ip_address_ranges`


ndellosa95:
### Description

astro_deployment.airflow3_deployment: Creating...
╷
│ Error: Provider produced inconsistent result after apply
│ 
│ When applying changes to astro_deployment.airflow3_deployment, provider
│ "provider[\"registry.terraform.io/astronomer/astro\"]" produced an
│ unexpected new value: .remote_execution.allowed_ip_address_ranges: was
│ null, but now cty.SetValEmpty(cty.String).
│ 
│ This is a bug in the provider, which should be reported in the provider's
│ own issue tracker.

### Steps To Reproduce

Try and create a remote execution deployment without `allowed_ip_address_ranges` provided. The provider lists this parameter as optional [here](https://registry.terraform.io/providers/astronomer/astro/latest/docs/resources/deployment#allowed_ip_address_ranges-2).

### Acceptance Criteria

- This parameter should actually be optional

Poornima030596:
same issue for deployment resource when updating hibernation schedule.

Error: Provider produced inconsistent result after apply
│
│ When applying changes to
│ astro_deployment.deployment["project6"], provider 
│ "provider[\"registry.terraform.io/astronomer/astro\"]"
│ produced an unexpected new value: .updated_at:    
│ was cty.StringVal("2026-01-05 07:21:29.989 +0000  
│ UTC"), but now cty.StringVal("2026-01-05
│ 07:23:42.829 +0000 UTC").
│
│ This is a bug in the provider, which should be    
│ reported in the provider's own issue tracker.     
╵

chobberoni:
closed as the core code is merged -- RELEASED LAST WEEK https://github.com/astronomer/terraform-provider-astro/releases/tag/v1.2.6

PR #314, opened 2026-05-12
Fix inconsistent result for remote_execution.allowed_ip_address_ranges CPP-527

bote795:
## Description

When creating a deployment with `remote_execution.enabled = true` and the
`allowed_ip_address_ranges` field **omitted** from HCL, Terraform fails with:

```
Error: Provider produced inconsistent result after apply

When applying changes to astro_deployment.<name>, provider
"provider[\"registry.terraform.io/astronomer/astro\"]" produced an unexpected
new value: .remote_execution.allowed_ip_address_ranges: was null, but now
cty.SetValEmpty(cty.String).
```

A second `terraform apply` settles the state, but the first apply error blocks
users from creating remote-execution deployments via Terraform unless they
explicitly set `allowed_ip_address_ranges`.

### Root cause

The Astro API always returns `allowedIpAddressRanges` as an array (empty `[]`
when unset rather than `null`). The provider, however, sent the field via
`*[]string` with `omitempty` and declared the schema attribute `Optional`-only.

So when a user omitted the field in HCL:

- Terraform plan: `allowed_ip_address_ranges = null`
- API stores `nil`, returns `[]`
- Provider sets state to empty set
- Plan (`null`) ≠ state (`[]`) → Terraform rejects the apply

### Fix

Align the resource schema with the API contract. Mark
`remote_execution.allowed_ip_address_ranges` as `Computed` with an empty-set
default and add `UseStateForUnknown` for updates. Plan and post-apply state
now both resolve to `[]` when the user omits the field.

```go
"allowed_ip_address_ranges": resourceSchema.SetAttribute{
    MarkdownDescription: "The allowed IP address ranges for remote execution",
    Optional:            true,
    Computed:            true,
    ElementType:         types.StringType,
    Default: setdefault.StaticValue(
        types.SetValueMust(types.StringType, []attr.Value{}),
    ),
    PlanModifiers: []planmodifier.Set{
        setplanmodifier.UseStateForUnknown(),
    },
},
```

**Behavior note for users:** This fix codifies the API’s actual semantics for allowed_ip_address_ranges: the API always returns an array (never null), and once a deployment exists, the field has a value. With UseStateForUnknown, removing the field from HCL after it was set is a no-op (prior value is preserved). To clear, set explicitly to []. Previously, both operations triggered the inconsistent-result error — there was no working “remove to clear” path, so this isn’t a behavior regression

## 🎟 Issue(s)

Fixes internal astronomer issue

## 🧪 Functional Testing

Tested end-to-end using a
`dev_overrides` build of the branch, with the same exact reproduction config
from #244.

### Reproduction (before the fix) — confirms the bug exists on `main`

Built the provider from `main` (commit `366d28e`) and ran:

```hcl
resource "astro_deployment" "bug_repro" {
  name                  = "cpp527-bug-repro"
  type                  = "DEDICATED"
  cluster_id            = "<dedicated-cluster-id>"
  workspace_id          = "<workspace-id>"
  contact_emails        = []
  executor              = "ASTRO"
  is_cicd_enforced      = true
  is_dag_deploy_enabled = false
  is_development_mode   = false
  is_high_availability  = false
  scheduler_size        = "SMALL"
  environment_variables = []
  remote_execution = {
    enabled = true
    # allowed_ip_address_ranges intentionally omitted
  }
}
```

Result:

```
astro_deployment.bug_repro: Creating...

Error: Provider produced inconsistent result after apply

When applying changes to astro_deployment.bug_repro, provider
"provider[\"registry.terraform.io/astronomer/astro\"]" produced an unexpected
new value: .remote_execution.allowed_ip_address_ranges: was null, but now
cty.SetValEmpty(cty.String).
```

### Verification (after the fix) — same HCL, same dev environment

Built the provider from this branch and ran the same apply.

Plan diff for the `remote_execution` block now shows the default:

```
+ remote_execution = {
    + allowed_ip_address_ranges = []
    + enabled                   = true
    + remote_api_url            = (known after apply)
  }
```

Apply output:

```
astro_deployment.cpp527_probe: Creating...
astro_deployment.cpp527_probe: Creation complete after 2s [id=cmp2wqygn000101kivb2aji0t]

Apply complete! Resources: 1 added, 0 changed, 0 destroyed.
```

Second `terraform plan` confirms idempotency:

```
No changes. Your infrastructure matches the configuration.
Terraform has compared your real infrastructure against your configuration and found no differences, so no changes are needed.
```

### Both runs visible on dev via `astro deployment list`

```
 NAME                 CLUSTER           CLOUD PROVIDER     REGION      DEPLOYMENT ID                 DEPLOYMENT TYPE     REMOTE EXECUTION
 cpp527-probe         testTerraform     AZURE              westus2     cmp2wqygn000101kivb2aji0t     DEDICATED           true
 cpp527-bug-repro     testTerraform     AZURE              westus2     cmp2x5eet00af01lmbhkvba3q     DEDICATED           true
```

`cpp527-probe` was created cleanly by the fixed provider. `cpp527-bug-repro`
is the orphan created by the buggy provider on `main` — Astro accepted the
API call and created the deployment, but Terraform rejected the response with
the inconsistent-result error before recording it in state. 

### Side-by-side

| Provider | Plan diff for `allowed_ip_address_ranges` | Apply result |
|---|---|---|
| `main` (pre-fix) | line absent (planned as `null`) | ❌ Provider produced inconsistent result after apply |
| This branch | `+ allowed_ip_address_ranges = []` | ✅ created in 2s, idempotent on replan |

### Acceptance test

Added `TestAcc_ResourceDeploymentRemoteExecutionAllowedIPRanges` in
`internal/provider/resources/resource_deployment_test.go`:

- **Step 1:** create a remote-execution deployment with
  `allowed_ip_address_ranges` omitted — asserts the apply succeeds in one
  shot and the implicit post-apply plan is empty.
- **Step 2:** update the same deployment to a multi-element list
  (`["10.0.0.0/8", "192.168.1.0/24"]`) — asserts both values round-trip
  faithfully and plan is empty.

Existing acceptance tests that exercise the field
(`TestAcc_ResourceDeploymentDedicated`, AWS/Azure/GCP cluster suites) updated
to pass the IP list explicitly through the test helper rather than relying on
the previous helper's hard-coded default.

## 📸 Screenshots

N/A — schema-level fix with no UI changes.

## 📋 Checklist

- [ ] Added/updated examples in the `examples/` directory _(existing examples already cover this field)_
- [x] Updated any related [documentation](https://github.com/astronomer/docs/) _(provider docs regenerated via `go generate`)_

### DIFF
--- a/internal/provider/resources/resource_cluster_test.go
+++ b/internal/provider/resources/resource_cluster_test.go
@@ -78,12 +78,13 @@ func TestAcc_ResourceClusterAwsWithDedicatedDeployments(t *testing.T) {
 						SchedulerSize: "SMALL",
 					}) +
 					dedicatedRemoteDeployment(dedicatedRemoteDeploymentInput{
-						ClusterId:     awsResourceVarId,
-						WorkspaceId:   workspaceResourceVarId,
-						Name:          awsDeploymentName,
-						Description:   "deployment description",
-						Executor:      "ASTRO",
-						TaskLogBucket: "s3://my-task-log-bucket",
+						ClusterId:              awsResourceVarId,
+						WorkspaceId:            workspaceResourceVarId,
+						Name:                   awsDeploymentName,
+						Description:            "deployment description",
+						Executor:               "ASTRO",
+						TaskLogBucket:          "s3://my-task-log-bucket",
+						AllowedIpAddressRanges: []string{"8.8.8.8/32"},
 					}),
 				Check: resource.ComposeTestCheckFunc(
 					// Check cluster
@@ -290,12 +291,13 @@ func TestAcc_ResourceClusterAzureWithDedicatedDeployments(t *testing.T) {
 						SchedulerSize: "SMALL",
 					}) +
 					dedicatedRemoteDeployment(dedicatedRemoteDeploymentInput{
-						ClusterId:         azureResourceVarId,
-						WorkspaceId:       workspaceResourceVarId,
-						Name:              azureDeploymentName,
-						Description:       utils.TestResourceDescription,
-						Executor:          "ASTRO",
-						TaskLogUrlPattern: "https://s3.amazonaws.com/my-task-log-bucket/{{ task_id }}/{{ execution_date }}/{{ filename }}",
+						ClusterId:              azureResourceVarId,
+						WorkspaceId:            workspaceResourceVarId,
+						Name:                   azureDeploymentName,
+						Description:            utils.TestResourceDescription,
+						Executor:               "ASTRO",
+						TaskLogUrlPattern:      "https://s3.amazonaws.com/my-task-log-bucket/{{ task_id }}/{{ execution_date }}/{{ filename }}",
+						AllowedIpAddressRanges: []string{"8.8.8.8/32"},
 					}),
 				Check: resource.ComposeTestCheckFunc(
 					// Check cluster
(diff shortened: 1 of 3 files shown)


aliotta (approved):

chobberoni:
https://linear.app/astronomer/issue/CPP-527/bug-inconsistent-result-after-apply-for-remote-executionallowed-ip

PR #317, opened 2026-05-27
fix: improve astro_users performance by using list instead of set

fboissier:
## Description

This pull request improves performance of the astro_users data source by changing the users attribute from a nested Set to a nested List.

Changes included:
- Updated users schema from SetNestedAttribute to ListNestedAttribute
- Updated users model type from types.Set to types.List and value construction from SetValue to ListValue

Rationale:
- Nested sets of complex objects are expensive to hash and compare during provider framework validation and state encoding.
- A list significantly reduces this overhead for large user payloads.

Performance impact observed:
- terraform plan execution for astro_users improved from about 30 seconds to about 1 second for an organization with approximately 1000 users.

## 🎟 Issue(s)
N/A

## 🧪 Functional Testing

1. Built and loaded the local provider through dev override.
2. Ran terraform init and terraform plan against the astro-test workspace.
3. Verified astro_users resolves correctly with the updated schema and model.
4. Compared behavior before and after to confirm reduced plan latency in practice.

## 📸 Screenshots

No screenshots for this backend/provider change.

## 📋 Checklist

### DIFF
--- a/docs/data-sources/users.md
+++ b/docs/data-sources/users.md
@@ -39,7 +39,7 @@ output "example_users" {
 
 ### Read-Only
 
-- `users` (Attributes Set) (see [below for nested schema](#nestedatt--users))
+- `users` (Attributes List) (see [below for nested schema](#nestedatt--users))
 
 <a id="nestedatt--users"></a>
 ### Nested Schema for `users`
--- a/internal/provider/models/users.go
+++ b/internal/provider/models/users.go
@@ -12,7 +12,7 @@ import (
 
 // Users describes the data source data model.
 type Users struct {
-	Users        types.Set    `tfsdk:"users"`
+	Users        types.List   `tfsdk:"users"`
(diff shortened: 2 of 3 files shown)


AmineTay (approved):
review ok

bote795:
Opened #319 which builds on this branch and adds an explicit `id:asc` sort to the `ListUsers` call.

Rationale: switching `users` from a `Set` to an ordered `List` makes the API's return order load-bearing. Without a stable sort, two things surface:
- **Spurious plan diffs** when the default ordering shifts between reads (a `List` reports `~` on every shifted index; a `Set` hid this).
- **Pagination boundary races** — the `Read` loop paginates with `offset += 1000` and no sort, so a changing order can skip/duplicate users across pages.

Sorting by `id` (immutable, unique) makes both pagination and list ordering deterministic in one line.

PR #321, opened 2026-06-26
fix: oapi-codegen discriminator injection for optional union fields

aliotta:
## Description

oapi-codegen's upstream `union.tmpl` generates `From*`/`Merge*` helpers that assign the discriminator with `v.<Field> = "VALUE"`. That fails to compile when the variant's discriminator field is not in the schema's `required` list — oapi-codegen generates the field as a typed pointer (e.g. `*UpdateDagDurationAlertRequestType`), and a bare string literal can't be assigned to it.

This blocked regenerating the platform/IAM clients against the latest core spec, which now exposes optional-discriminator union variants under `UpdateAlertRequest` (`UpdateDagDurationAlertRequest`, `UpdateDagFailureAlertRequest`, etc. — none of which list `type` as required).

### Fix

Override `union.tmpl` so `From*`/`Merge*` inject the discriminator into the marshaled JSON instead of writing it directly to the variant struct:

```go
b, err := json.Marshal(v)
// ...
obj := make(map[string]json.RawMessage)
json.Unmarshal(b, &obj)
obj["type"] = json.RawMessage(`"DAG_DURATION"`)
b, err = json.Marshal(obj)
t.union = b
```

The new code path is type-agnostic — it works whether the variant's discriminator field is required (value type) or optional (pointer type), so the same template handles both `Create*AlertRequest` (required) and `Update*AlertRequest` (optional) variants identically.

Override is dropped at `internal/clients/oapi-templates/union.tmpl` and wired into `api_client_gen` via oapi-codegen's `-templates` flag.

## 🎟 Issue(s)

N/A — surfaced while regenerating clients against the latest core spec.

## 🧪 Functional Testing

- [x] `make api_client_gen` succeeds end-to-end
- [x] `go build ./...` clean across the whole provider module
- [x] `go test ./...` passes all suites
- [x] Generated `From*`/`Merge*` helpers compile for both required-discriminator (`Create*AlertRequest`) and optional-discriminator (`Update*AlertRequest`) variants

## 📋 Checklist

- [x] Added/updated applicable tests — existing test suites cover round-tripping; no new test added for codegen output itself (covered by build + existing union round-trip tests)
- [ ] Added/updated examples in the `examples/` directory — N/A
- [ ] Updated any related [documentation](https://github.com/astronomer/docs/) — N/A


bote795:
A few follow-ups

## 🟡 #1 — `ValueByDiscriminator()` is broken for notification-channel unions (latent regression)

The regen changed `CreateNotificationChannelRequest`/`UpdateNotificationChannelRequest.ValueByDiscriminator()` so the switch cases are **Go type names** instead of the **wire discriminator values**:

```go
// before (main):  case "DAG_TRIGGER":
// after  (here):  case "CreateDagTriggerNotificationChannelRequest":
```

`Discriminator()` still reads the wire `type` field (`"DAG_TRIGGER"`), so the switch can never match → it always returns `unknown discriminator value: DAG_TRIGGER`.

Proven with one test run identically against both branches:

| Branch | `Discriminator()` | `ValueByDiscriminator()` |
|--------|-------------------|--------------------------|
| `main` | `"DAG_TRIGGER"` | ✅ resolves to `CreateDagTriggerNotificationChannelRequest` |
| this PR | `"DAG_TRIGGER"` | ❌ `error: unknown discriminator value: DAG_TRIGGER` |

**Severity is P2, not blocking, because it's latent** — `ValueByDiscriminator` has zero callers outside generated code today. But it's a real regression on the client's public surface and a footgun for anything that calls it later.

**Root cause is the core spec, not this template.** The notification-channel `oneOf` discriminator `mapping` is now identity-keyed (keys = schema/type names) rather than value-keyed (`DAG_TRIGGER -> …Request`). The alert/cluster/deployment unions kept value-keyed mappings and are fine. Fix is to correct the mapping in the core OpenAPI spec and regen — worth flagging to the API team.

> Note: the same root cause makes the `(ne $value $element)` guard skip discriminator injection for these channels, but that's **not** a wire bug — Create variants have a required `Type` field and Update variants are called with `Type: &channelType`, so the marshaled JSON always carries `"type"`. Verified.

## 🟡 #2 — Forked `union.tmpl` has no provenance (`union.tmpl:1`)

This is a hand-modified fork of oapi-codegen's `union.tmpl`, but nothing records which upstream version it forked from or that it's customized. On the next `DESIRED_OAPI_CODEGEN_VERSION` bump, whoever regenerates won't know the injection blocks + `(ne $value $element)` guard must be re-applied. Suggest:
- Header comment in `union.tmpl` (forked-from version + rationale: inject discriminator via JSON for optional-pointer fields).
- One-line comment on the `-templates` flag in the Makefile.
- An "API client regeneration" note in `CONTRIBUTING.md`.


## 🟢 #3 — Missing trailing newline (`Makefile:101`)

Both regenerated `oapi-codegen` lines end with `\ No newline at end of file`. Trivial — add a trailing newline.

## 🔵 #5–#7 — Test coverage (advisory)

- **Regression test for discriminator serialization** — none exists. A table-driven test asserting `From*` output contains the right `"type"` for each variant (especially the `Update*Alert*` optional-pointer family this PR fixes) would lock in the fix.
- **Codegen-drift CI check** — no `make api_client_gen` drift gate, so a future oapi-codegen upgrade could silently drop this override and re-break optional-pointer compilation undetected.
- A round-trip test through `ValueByDiscriminator()` would have caught #1 immediately.

bote795 (approved):

PR #325, opened 2026-07-07
Add Bulk alerts operations from Labs API

dwillson:
## Description

Adds a new astro_alerts collection resource that manages many alerts as a single Terraform resource, backed by Core's labs bulk alert API. It batches create/update/delete calls and auto-chunks requests that exceed the per-request limits (30 create/update, 20 delete), so a config can declare any number of alerts in one block.

This complements the existing single-alert astro_alert resource (unchanged in behavior; shared request-building logic was extracted so both can reuse it).

What's included

* New resource astro_alerts — alerts keyed by a stable user-defined string; CRUD diffs by key and issues chunked bulk calls (internal/provider/resources/resource_alerts.go, schemas/alerts_resource.go, models/alerts_resource.go).
* Labs API client — generated client for the labs bulk endpoints (internal/clients/labs/), wired into the provider (provider.go, models/api_clients.go) and the codegen Makefile.
* Shared request builders — per-type alert request construction factored out of resource_alert.go into alert_requests.go / labs_alert_requests.go (net reduction in resource_alert.go).
* Tests — chunking unit test (alerts_chunk_test.go), resource model test (alerts_resource_test.go), and a labs discriminator-serialization regression test (internal/clients/labs/discriminator_test.go).
* Docs & examples — docs/resources/alerts.md, examples/resources/astro_alerts/, and a CONTRIBUTING note on client generation.

## 🧪 Functional Testing

- go build ./..., go test ./internal/... pass; make api_client_gen regenerates all clients cleanly (no drift).
- Manual end-to-end against a local Core, via a Terraform config exercising all paths:
- Bulk create of a varied set (multiple alert types / severities / notification channels).
- Create of 35 alerts in one resource → 2 auto-chunked create requests.
- Update (in-place) and bulk delete of 30 alerts → 2 auto-chunked delete requests.
- terraform plan reports no drift after each apply (read round-trip verified).

## 📋 Checklist


aliotta (approved):

aliotta:
We should make sure we add some documentation around telling people that they should not put the same alert in both alert and alerts resources. Looks like terraform supports deprecation notes.

e.g.
 func (d *usersDataSource) Schema(ctx context.Context, _ datasource.SchemaRequest, resp *datasource.SchemaResponse) {
      resp.Schema = schema.Schema{
          DeprecationMessage: "Use `astro_users_list` instead. `astro_users` returns users as a Set which causes slow terraform plan on large orgs and will be removed in v2.0.",
          Attributes: ...,
      }
  }
  
  Terraform surfaces this as a warning diagnostic on every plan/apply referencing the data source. It also shows up in the generated docs (tfplugindocs adds a "Deprecated" banner).
  
  
  So lets add one for the alert data store

aliotta (approved):

PR #328, opened 2026-07-17
Add astro_allowed_ip_address_ranges resource (CPP-894)

### LINKED ISSUE #264: Support for setting IP Access List from terraform
### Description

We would like the ability to set the ip access list from terraform.

https://www.astronomer.io/docs/astro/ip-access-list

### Acceptance Criteria

- Access list managed by terraform
- (Nice to have) a backend check and API rejection of the access list if the request would result in terraform losing access to astronomer

### Anything else?

_No response_

coleheflin:
## Description

Adds an `astro_allowed_ip_address_ranges` resource that manages an organization's IP access list from Terraform (fixes #264). The resource is authoritative: `ip_address_ranges` is the full set of CIDR ranges, and any ranges not present are removed on apply.

- **Writes** go through the new labs bulk endpoints:
  - `POST /labs/v1/organizations/{organizationId}/allowed-ip-address-ranges` (bulk create, atomic, up to 1000 CIDRs per request)
  - `DELETE /labs/v1/organizations/{organizationId}/allowed-ip-address-ranges` (bulk delete by ID)
- **Reads** list the existing `GET /iam/v1beta1/organizations/{organizationId}/allowed-ip-address-ranges` collection (paginated) — labs has no list endpoint. Create/update diff the plan against the live list to decide which CIDRs to add or remove. Update creates before deleting so the machine running Terraform never loses coverage mid-apply. Read-via-list mirrors the pattern from `astro_alerts` (#325).
- **Import** is supported (singleton keyed by organization ID), so an existing access list can be adopted with `terraform import` instead of re-declaring it.

> ⚠️ Adding the first range turns enforcement on: that first apply is rejected unless the submitted ranges include the public IP of the machine running Terraform. Documented on the resource.

## Note on generated client code

- **Labs create/delete are generated.** `make api_client_gen` now includes the `AllowedIpAddressRange` tag for the labs client (see `Makefile`), producing the additions in `internal/clients/labs/api.gen.go`.
- **The iam list binding is hand-authored** in `internal/clients/iam/allowed_ip_address_ranges.go`. Adding `AllowedIpAddressRange` to the iam `include-tags` triggers an oapi-codegen v2.1.0 constant-naming collision that renames the unrelated `ListUsersParamsSorts` enum and breaks `data_source_users_list.go`, so the list binding is maintained by hand (matching the generated shape) until that's resolved upstream. A `Makefile` note records this.

## Testing

- `go build ./...`, `go vet ./...`, and `go test ./internal/...` pass — including unit tests for the create/delete diffing logic (`allowed_ip_address_ranges_diff_test.go`).
- `go generate ./...` regenerates docs/examples cleanly (`docs/resources/allowed_ip_address_ranges.md`).
- Added an opt-in acceptance test (`TestAcc_ResourceAllowedIpAddressRanges`) covering create, update (create-before-delete), and import. It's self-guarding: it skips unless `ACC_TEST_RUNNER_IP_CIDR` is set and keeps that CIDR in every config so the runner can't be locked out.
- Live end-to-end run is pending an environment with the labs routes deployed (dev today, prod next week) — run instructions are in the comment thread.


coleheflin:
@bote795 — following up on this PR. Since the last push I added three hardening changes plus an acceptance test, and I'm hoping you can run the acceptance test against **dev** (where the labs `allowed-ip-address-ranges` routes are deployed.

### What changed
- **`Update` now creates before deleting.** IP enforcement is per-request and the server-side guard only covers the empty→first-entry transition (not deletes), so delete-before-create could remove the caller's covering range and 403 the rest of the apply. Create-first keeps the runner covered throughout.
- **Added `ImportState`** (computed `id` = organization ID) so orgs with an existing access list can `terraform import` instead of re-declaring (which would 409 / produce an inconsistent-result error on first apply).
- **Corrected the docs**: the API doesn't reject lockout-inducing changes on a non-empty list — the real gotcha is the *first* apply 400s unless the submitted ranges include the runner's public IP. Warn accordingly + document import-first.

### Ask: run the acceptance test against dev
The test (`TestAcc_ResourceAllowedIpAddressRanges`) is **opt-in and self-guarding** — it skips unless `ACC_TEST_RUNNER_IP_CIDR` is set, and it keeps that CIDR in every applied config so the runner is never locked out. It covers create (empty→first), update via create-before-delete, and import by org ID.

Please run it against a **disposable/test org on dev**, since it mutates that org's real IP access list:

```bash
export ASTRO_API_HOST="<dev api host>"
export HOSTED_ORGANIZATION_ID="<disposable dev org id>"
export HOSTED_ORGANIZATION_API_TOKEN="<org token for that org>"
# CIDR that covers the machine running the test (your public egress IP). Kept in every
# applied config so the test never locks the runner out:
export ACC_TEST_RUNNER_IP_CIDR="$(curl -s https://api.ipify.org)/32"

TF_ACC=1 go test ./internal/provider/resources/ -run TestAcc_ResourceAllowedIpAddressRanges -v
```

Notes:
- `TestAccPreCheck` (shared by all acc tests) also expects the standard `HYBRID_*` / `HOSTED_*` vars to be set even for this single test — your usual acc-test env should cover them.
- On dev egress: if the dev ingress doesn't forward a usable `X-Forwarded-For`, the first-range create will 400 by design (that's the guard) — worth confirming the runner's IP actually lands in `X-Forwarded-For` on dev.

If it's easier for you to smoke-test by hand instead: `terraform apply` a single range covering your IP → `terraform plan` (expect no drift) → `terraform import` on a second config → `terraform destroy`. Thanks!

bote795 (commented):
## Code review — `astro_allowed_ip_address_ranges`

Careful, well-documented PR. The lockout reasoning, create-before-delete ordering, empty-set handling, and pagination termination are all sound. Findings cluster around one root cause (post-apply state sourced from the API re-list) plus retry-idempotency and input-validation hardening.

### Themes
- **CIDR canonicalization / plan-state consistency** (#1, #4, #7) — one decision resolves the group.
- **Retry idempotency after a partial write** (#2, #5).
- **Input validation** (#3).

---

### P1

**#1 — Create/Update store the API re-list into a `Required` attribute; a non-canonical CIDR triggers "Provider produced inconsistent result after apply."**
`resource_allowed_ip_address_ranges.go` Create (`data.IpAddressRanges = setVal`) and Update (`plan.IpAddressRanges = setVal`), where `setVal` comes from `r.listAll(ctx)`, not the plan. `ip_address_ranges` is `Required` with no plan modifiers, and Terraform core enforces `newState == plannedValue` for a fully-known non-computed attribute. If the API returns a CIDR in a different string form than submitted — host bits (`10.1.2.3/8` → `10.0.0.0/8`), IPv6 compaction/case — the applied set differs from the plan and the apply aborts.

This is the same class already fixed in this repo: #314 (`remote_execution.allowed_ip_address_ranges`, `cdf3f89`) and #244. The import-first mitigation covers the *pre-existing-ranges* trigger but not canonicalization.

**Suggestion:** canonicalize CIDRs to one form before both diffing and storing state, or store the plan value in Create/Update and rely on Read for drift.

---

### P2

**#2 — Non-idempotent create: a transient failure on the `listAll` that follows a successful `bulkCreate` strands ranges server-side with nothing in state; the retry 409s.**
`State.Set` runs only after both `bulkCreate` and `listAll` succeed. If `listAll` errors after create committed, the resource is never written to state; the next apply re-runs Create → `LabsCreate...` 409 (the generated `JSON409` confirms duplicate-create conflicts). Reachable without the >1000 path.
**Suggestion:** treat 409/duplicate create as success, or persist `data.Id` before the post-create list so retry enters Update; at minimum surface a diagnostic.

**#3 — No client-side CIDR validator, despite in-repo precedent.**
`ip_address_ranges` is a `Required` set of strings with no `Validators`, though the description says "in CIDR format" and a bad value is only caught by a `400` partway through a destructive apply. The repo already does this in `schemas/deployment.go` via `setvalidator.ValueStringsAre(stringvalidator.RegexMatches(...))`, and there's an `internal/provider/validators` package.
**Suggestion:** add a CIDR set validator so malformed input fails at plan time.

---

### P3

**#4 — `diffCidrs` never converges under canonicalization (downstream of #1).** Exact string set-diff: if state holds the canonical form and config the non-canonical form, every plan sees them as disjoint → re-create + delete each apply, and #1 then crashes it. Resolved by the #1 decision.

**#5 — Multi-chunk (>1000) partial write leaves state-vs-reality divergence.** In `bulkCreate`, if a later chunk fails after an earlier one committed, Create strands ranges (see #2) and Update returns before deletes run; retry 409s on the committed chunk. Rare, but the same idempotency trap as #2.

**#6 — An OK-status non-JSON/empty body silently empties the authoritative set.** `if listResp.JSON200 == nil { break }` — `NormalizeAPIError` passes 200/204, and the parser only fills `JSON200` on `200 + Content-Type: json`. A `200 text/html` proxy interstitial (or 204) yields an empty list, and Read overwrites state to empty → next plan proposes deleting every managed range and 409s recreating.
**Suggestion:** error on a normalized-OK response with `JSON200 == nil` rather than returning empty.

**#7 — `idsForCidrs` silently skips CIDRs it can't resolve (authoritative under-delete, no diagnostic).** `if id, ok := byCidr[c]; ok { ids = append(ids, id) }` with no else. On destroy/update-delete, a state CIDR missing from the live list (normalization drift, or #6) is silently not deleted; the apply reports success while the range persists on a security allowlist.
**Suggestion:** warn (or error) on an unresolved CIDR.

**#9 — Hand binding omits the method from `iam.ClientWithResponsesInterface`.** Generated code always adds `...WithResponse` to the interface; this adds it only to the concrete type. It compiles, but diverges from generated shape and blocks mock-based testing.
**Suggestion:** add it to the interface too (also unblocks the mocked tests below).

**#10 — Typo:** "coegen" → "codegen" in the `iam/allowed_ip_address_ranges.go` header comment.

---

### Testing gaps
- The acceptance test always `t.Skip`s in CI (`ACC_TEST_RUNNER_IP_CIDR` is unset in `.github/workflows/testacc.yml`), so the only test running in CI is `TestUnit_diffCidrs` — the safety-critical create-before-delete ordering is never exercised automatically.
- `listAllRanges` pagination (multi-page, empty-page, `TotalCount==0` early stop) and `bulkCreate`/`bulkDelete` chunking are untested and not mockable as written (resource holds concrete client types; no labs mock). The pattern in `internal/provider/common/role_test.go` applies once #9 + a labs mock land.
- `diffCidrs` coverage is good; only `nil,nil` and intra-slice duplicates are missing (low value).

### Residual risks (not defects)
- Lockout on narrowing/replacing the covering range is inherent to an authoritative allowlist and is explicitly documented. Create-before-delete correctly eliminates the transient gap; the empty→first-entry `400` guard is correctly surfaced.

---

**Verdict: Ready with fixes.** Solid PR. #1 is a decision to make before merge; #2 and #3 are strongly recommended; the rest are hardening.

coleheflin:
Closing as a duplicate — the `astro_allowed_ip_address_ranges` resource was implemented and merged in #330 (released in v1.5.0), which supersedes this branch. #330 incorporates the same hardening this PR was addressing (persist planned values for plan/state consistency, plan-time `net.ParseCIDR` validation, safe reads, unresolved-CIDR warning) plus idempotent 409-as-success handling for partial multi-chunk writes. No unique feature work is lost by closing this.

PR #335, opened 2026-08-14
fail fast during tf plan state during the update of custom roles associated with a Direct access token

sakethsomaraju:
## Description

Currently adding a permission to an existing deployment-scoped `astro_custom_role` passed `terraform plan` cleanly but failed at `terraform apply`, with no indication in the plan that anything was wrong. Root
  cause: the role had one or more Direct Access Tokens (DATs) attached, and DATs snapshot their role's permissions at creation time, the API rejects permission changes to a role while any DAT references it. Nothing in local state/config
  diffing can see this, so it silently passed plan and only surfaced as a generic 400 at apply.

  This PR:

  - Fails fast at plan time. Adds a `ModifyPlan` check on `astro_custom_role`: when `permissions` changes, the provider looks up whether any Direct Access Token references the role (by name, since token role assignments are keyed by role
  name, not ID) and raises a clear plan-time error naming the conflicting token(s) and the required workaround, instead of letting a broken plan through. The API's own apply-time rejection is left untouched as the authoritative backstop, since
  state can still drift between plan and apply.
  - Surfaces token kind. Adds `kind` (`STANDARD` / `DIRECT_ACCESS`) to the `astro_api_token` resource, the `astro_api_token` data source, and the `astro_api_tokens` list data source. The IAM API already returns this field, but the provider
  was silently dropping it, making it impossible to tell a DAT apart from a standard token via Terraform.

  Note: Direct Access Tokens themselves can't be created via `astro_api_token` (the API only allows creating them outside Terraform, e.g. via `astro deployment token` or the UI), so this is a best-effort provider-side safety net, not full
  lifecycle management of DATs.

## 🎟 Issue(s)
 CPP-940

## 🧪 Functional Testing

1. Created a custom role.
2. Attached it to a deployment scoped DAT.
3. Tried to add another granular permission to the role.
4. After fix, TF plan failed as expected.

## 📸 Screenshots

image

## 📋 Checklist

### DIFF
--- a/docs/data-sources/api_token.md
+++ b/docs/data-sources/api_token.md
@@ -37,6 +37,7 @@ output "api_token" {
 - `description` (String) API Token description
 - `end_at` (String) time when the API token will expire in UTC
 - `expiry_period_in_days` (Number) API Token expiry period in days
+- `kind` (String) API Token kind (STANDARD or DIRECT_ACCESS)
 - `last_used_at` (String) API Token last used timestamp
 - `name` (String) API Token name
 - `roles` (Attributes Set) The roles assigned to the API Token (see [below for nested schema](#nestedatt--roles))
--- a/docs/data-sources/api_tokens.md
+++ b/docs/data-sources/api_tokens.md
@@ -60,6 +60,7 @@ Read-Only:
 - `description` (String) API Token description
 - `end_at` (String) time when the API token will expire in UTC
 - `expiry_period_in_days` (Number) API Token expiry period in days
+- `kind` (String) API Token kind (STANDARD or DIRECT_ACCESS)
 - `last_used_at` (String) API Token last used timestamp
 - `name` (String) API Token name
 - `roles` (Attributes Set) The roles assigned to the API Token (see [below for nested schema](#nestedatt--api_tokens--roles))
--- a/docs/resources/api_token.md
+++ b/docs/resources/api_token.md
@@ -170,6 +170,7 @@ resource "astro_api_token" "imported_api_token" {
 - `created_by` (Attributes) API Token creator (see [below for nested schema](#nestedatt--created_by))
 - `end_at` (String) time when the API token will expire in UTC
 - `id` (String) API Token identifier
+- `kind` (String) API Token kind (STANDARD or DIRECT_ACCESS). This resource always creates STANDARD tokens.
 - `last_used_at` (String) API Token last used timestamp
 - `short_token` (String) API Token short token
 - `start_at` (String) time when the API token will become valid in UTC
--- a/go.mod
+++ b/go.mod
@@ -21,6 +21,7 @@ require (
 	github.com/samber/lo v1.39.0
 	github.com/stretchr/testify v1.9.0
 	golang.org/x/exp v0.0.0-20240719175910-8a7402abbf56
+	golang.org/x/sync v0.8.0
 )
 
 require (
@@ -90,7 +91,6 @@ require (
 	golang.org/x/crypto v0.26.0 // indirect
 	golang.org/x/mod v0.20.0 // indirect
 	golang.org/x/net v0.28.0 // indirect
-	golang.org/x/sync v0.8.0 // indirect
 	golang.org/x/sys v0.23.0 // indirect
 	golang.org/x/text v0.17.0 // indirect
 	golang.org/x/tools v0.24.0 // indirect
--- a/internal/provider/models/api_token.go
+++ b/internal/provider/models/api_token.go
@@ -17,6 +17,7 @@ type ApiTokenDataSource struct {
 	Description        types.String `tfsdk:"description"`
 	ShortToken         types.String `tfsdk:"short_token"`
 	Type               types.String `tfsdk:"type"`
+	Kind               types.String `tfsdk:"kind"`
 	StartAt            types.String `tfsdk:"start_at"`
 	EndAt              types.String `tfsdk:"end_at"`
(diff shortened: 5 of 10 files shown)


bote795:
## Code review — plan-time DAT guard + `kind` attribute

Reviewed the branch against `main` (merge-base `9ee76bb`). Happy path is solid and the `kind` wiring is consistent across the resource, both data sources, and the element-attribute map. Four things worth addressing, plus a few residual notes. Verdict: **ready with fixes.**

### P1

**1. Unbounded serial Direct Access Token fan-out at plan time, no timeout** — `internal/provider/resources/resource_custom_role.go:168`
`findDirectAccessTokensUsingRole` lists every DAT in the org (paginated) then issues one `GetApiTokenWithResponse` **per token, sequentially**, over a context with no deadline and a bare `&http.Client{}` (`internal/clients/iam/api.gen.go:1122`, no timeout). This runs on every `terraform plan` that changes a role's `permissions`. A large-DAT org or a slow IAM connection turns a plan into a long or indefinite hang.

Proposed fix (agreed direction):
- Wrap the lookup in `context.WithTimeout(ctx, <const budget>)` and pass the bounded ctx into both calls. Note this must be a **hardcoded const budget** — `ModifyPlan` can't use a `timeouts {}` block (it isn't a CRUD op), so it's not user-tunable. On timeout, fail open (same as any lookup error) — see #4 for visibility.
- Run the per-token GETs through `errgroup.SetLimit(8–16)` (`golang.org/x/sync` is already an indirect dep) instead of a sequential loop, collecting matches under a mutex.
- **No memoization.** A cache on the resource struct doesn't dedupe across roles (fresh instance per resource plan call); a provider/package-level cache would leak stale reads across plan operations if the process is reused. The number of roles changing in one plan is normally small, so parallelizing the per-token GETs is the real lever.
- Highest-leverage longer-term fix is API-side: if `ListApiTokens` returned `roles` or supported a role filter, the per-token fan-out disappears entirely. Worth raising with the IAM team.

**2. New guard logic has no unit-test coverage; only a live-HOSTED acceptance test exercises it** — `internal/provider/resources/resource_custom_role.go:110`
The only test touching `ModifyPlan`/`findDirectAccessTokensUsingRole` is `TestAcc_ResourceCustomRoleDirectAccessTokenConflict`, gated by `TestAccPreCheck` (requires `HOSTED_*` env), so a default no-network CI run exercises none of the new logic — fail-open branch, multi-page pagination, zero-conflict path, and the `>10`-token truncation are all unverified. Suggest a package-level unit test (`httptest.NewServer` + `iam` client, per `allowed_ip_address_ranges_consistency_test.go`) covering those branches.

**3. Acceptance test can leak the out-of-band DAT (and role) when the guard regresses** — `internal/provider/resources/resource_custom_role_test.go:213`
Step 2 uses `ExpectError`. If the guard ever regresses so the expected plan error doesn't appear, the plugin-testing harness fails that step and **skips step 3 — the token deletion**. `CheckDestroy` only checks role existence and can't delete the role while the token references it, so on the exact regression this test guards, it leaks a token + role into the shared HOSTED org. Suggest registering the token cleanup with `t.Cleanup(...)` right after `testAccCreateDirectAccessTokenForRole` succeeds (or deleting it in `CheckDestroy` before the role check) so it runs regardless of which step fails.

### P2

**4. Fail-open path is invisible in normal `terraform plan` output** — `internal/provider/resources/resource_custom_role.go:111`
On a lookup error the guard calls only `tflog.Warn` and returns (failing open is the right tradeoff, since apply-time API rejection is the backstop). But `tflog.Warn` is invisible without `TF_LOG`, so a transient IAM failure silently disables the safety net — undercutting the "fail fast / be visible" intent exactly when the dependency is degraded. Suggest also calling `resp.Diagnostics.AddWarning("Unable to verify Direct Access Token compatibility", <detail>)`.

### Residual notes (non-blocking)

- **Pagination advances by the requested limit, not the returned page size** (`resource_custom_role.go:304`): `offset += 1000`. If the IAM list endpoint ever caps a page below 1000, tokens get skipped and the guard silently misses a conflict. Termination is otherwise correct (one redundant empty request, no infinite loop). The safer idiom exists at `data_source_environment_objects.go:138`. Low risk; apply-time rejection is the backstop.
- **`kind` + `UseStateForUnknown` has no schema-version upgrade** (`schemas/api_token.go:120`): on pre-existing state (no `kind` key), a plan that skips refresh (`-refresh=false` or applying a saved plan) can show a one-time spurious in-place Update. The default refresh workflow repopulates `kind` in `Read()` first, so the common case is clean, and this matches the provider's existing convention (no resource implements `UpgradeState`). Worth a release-note mention.
- **Pagination loop duplicates `data_source_api_tokens.go`** — a shared `iam.ListAllApiTokens(...)` helper would remove the copy (optional cleanup).
- **Testing gaps:** `kind` value is never asserted in any resource/data-source test (`TestCheckResourceAttr(..., "kind", string(iam.ApiTokenKindSTANDARD))`); no coverage for multi-page pagination, large token counts, rename+permissions, or the no-conflict case.

### Suggested order
1. **#1** — bound + deadline + parallelize the fan-out.
2. **#3** — move DAT cleanup to `t.Cleanup`/`CheckDestroy`.
3. **#4** — `AddWarning` on the fail-open path.
4. **#2** — httptest unit tests for the guard.

<sub>All findings verified against source and passed an independent validation pass. The independent cross-model adversarial pass could not run this round (external-model quota), so these rest on single-model reviewers + validation — worth a second human skim on #1.</sub>

sakethsomaraju:
Addressed P1 and P2.

PR #336, opened 2026-08-31
Add astro_notification_channels bulk resource

bote795:
## What

Adds a new **`astro_notification_channels`** resource that manages many notification channels as a single resource, batching create/delete through the Labs bulk endpoints. 

 Terraform applies that create many `astro_notification_channel` resources hit the Core API one channel at a time and get rate limited. Batching sends one request per 30 channels instead of one per channel.

## How

- **Create / Delete** → Labs bulk endpoints (`LabsCreateNotificationChannels` / `LabsDeleteNotificationChannels`), auto-chunked at 30 per request. Created IDs are zipped back to map keys positionally, keeping a valid prefix on partial failure so Terraform never records the wrong id for a channel.
- **Update** → routes each changed channel through the single-channel `platform` `UpdateNotificationChannel` endpoint. Labs has **no bulk-update** endpoint for notification channels (unlike alerts). This keeps channel IDs stable (delete+recreate would change IDs and fail for channels referenced by an alert/monitor). Batching the update path is tracked in **CPP-968**.
- **Read** → refreshes via the platform `ListNotificationChannels` endpoint (Labs has no list), chunked at 100. Preserves the sensitive fields the API never returns (`webhook_url`, `api_key`, `integration_key`, `deployment_api_token`) from prior state, matching the singular resource.
- Regenerates the Labs client with the `NotificationChannels` tag. This prefixes the `AlertNotificationChannelType` constants (`EMAIL` → `AlertNotificationChannelTypeEMAIL`, etc.) because the bulk NC types reuse the same bare enum values; no provider code referenced the bare constants, so the rename is safe. Documented in the Makefile.

## Do not manage the same channel with both resources

Like `astro_alert` / `astro_alerts`, do not manage a given channel with both `astro_notification_channel` and `astro_notification_channels` — each claims ownership and they will churn against each other.

## Tests

- Unit: bulk-create short/missing-chunk abort + valid-prefix-on-failure, bulk-delete 404 tolerance, per-type request-builder union marshaling (SLACK/EMAIL/PAGERDUTY/OPSGENIE/DAG_TRIGGER).
- `go build ./...`, `go vet ./internal/...`, and unit suites for resources/models/schemas/clients all green. Docs generated via tfplugindocs.
- **Not yet run:** acceptance tests against a live org — validating against dev next.

🤖 Generated with [Claude Code](https://claude.com/claude-code)


internal/provider/resources/resource_notification_channels.go
```diff
@@ -0,0 +1,449 @@
...
+	}
+
+	// Labs has no bulk-update endpoint for notification channels; update each changed channel
+	// through the single-channel platform endpoint (CPP-968).
+	for _, k := range updateKeys {
+		id := stateElems[k].Id.ValueString()
+		if id == "" {
+			continue
+		}
+		if d := r.updateChannel(ctx, id, planElems[k]); d.HasError() {
```
dwillson, line 270: Appears to update every channel regardless of actual change. You should be able to skip if no changes detected (planElems[k]==stateElems[k])

internal/provider/resources/resource_notification_channels.go
```diff
@@ -0,0 +1,449 @@
...
+			writeDiags.Append(d...)
+		}
+	}
+
+	r.persistFromIds(ctx, &resp.State, &resp.Diagnostics, keyToId, planElems)
```
dwillson: If an update fails is the previous state being overwritten? Esp in the case of a sensitive-only change, the server side fails to update, but the changed version is being persisted

internal/provider/models/notification_channels_resource.go
```diff
@@ -0,0 +1,69 @@
...
+	"github.com/astronomer/terraform-provider-astro/internal/clients/platform"
+	"github.com/astronomer/terraform-provider-astro/internal/provider/schemas"
+	"github.com/hashicorp/terraform-plugin-framework/diag"
+	"github.com/hashicorp/terraform-plugin-framework/types"
+)
+
+// NotificationChannelsResource describes the astro_notification_channels (bulk) resource data model.
+// Channels are keyed by a stable user-defined string so Terraform can track each channel's identity
+// across applies while the resource batches the underlying API calls.
+type NotificationChannelsResource struct {
```
dwillson, line 15: Should the singular NotificationChannelResource get a deprecation notice with this change?

internal/provider/resources/resource_notification_channels.go
bote795: uppdated

dwillson (approved):

